PR #24, opened 2013-11-13
Load services dynamically

jamessharp:
This is a bit of a speculative PR

I've been integrating nerve/synapse into a [mesos](http://mesos.apache.org) cluster. Essentially I want to setup a cluster that I can distribute a number of services across and have automatic TCP routing such that if I declare that a service listens on port A, I can visit port A on any server in the cluster and get routed to an instance of the service, wherever it happens to live.

Combining nerve and synapse seems like a good choice for doing this, however I ran into a bit of a problem - you have to configure the synapse services up front. This was less than ideal because I don't want to have to go and reconfigure each instance of synapse individually every time I introduce a new service. Ideally what would happen is that synapse discovers that a new service has been deployed, it goes and fetches the config from somewhere and reloads itself with the new service.

This is what this PR attempts to do.

I've introduced two new components - a Service Discoverer and a Config Fetcher (not my finest hour when it comes to inventing names…). The Service Discoverer is responsible for discovering which services are in play and should be routed to. The Config Fetcher is responsible for fetching the config for new/unknown services.

At the moment I've just got a zookeeper discoverer and an S3 fetcher (the config is placed in the S3) but it looks like it is working quite well, and others can be added at any time.

Here is an example of a new `config.json` to enable this functionality

    {
      "services": {}
      "haproxy": {
        ...
      },
      "service_discoverer": {
        "method": "zookeeper",
        "hosts": [
          "10.1.2.2:2181",
          "10.1.2.3:2181",
          "10.1.2.4:2181"
        ],
        "path": "/smartstack/services",
        "fetcher": {
          "method": "s3",
          "credentials": "/opt/smartstack/aws_credentials.yml",
          "bucket": "service-config",
          "default_discovery": {
            "method": "zookeeper",
            "hosts": [
              "10.1.2.2:2181",
              "10.1.2.3:2181",
              "10.1.2.4:2181"
            ],
            "path": "/smartstack/services/%SERVICE_NAME%/services"
          }
        }
      }
    }

and here is a config file that would be stored on the S3 and fetched

    {
      "synapse": {
        "haproxy": {
          "port": 4400,
          "server_options": "check inter 2s rise 3 fall 2",
          "listen": [
            "mode http",
            "option httpchk /ping"
          ]
        }
      },
      "nerve": {
        "check_interval": 1,
        "checks": [
          {
            "type": "http",
            "uri": "/ping",
            "timeout": 1,
            "rise": 1,
            "fall": 1
          }
        ]
      }
    }

I'm not sure this code is production ready yet, but would be interested to know whether you'd be interested in having something like this in the core code?

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,17 +2,24 @@ PATH
   remote: .
   specs:
     synapse (0.6.0)
+      aws-sdk (~> 1.25.0)
       zk (~> 1.9.2)
 
 GEM
   remote: https://rubygems.org/
   specs:
+    aws-sdk (1.25.0)
+      json (~> 1.4)
+      nokogiri (>= 1.4.4, < 1.6.0)
+      uuidtools (~> 2.1)
     coderay (1.0.9)
     diff-lcs (1.2.4)
+    json (1.8.1)
     little-plugger (1.1.3)
     logging (1.7.2)
       little-plugger (>= 1.1.3)
     method_source (0.8.2)
+    nokogiri (1.5.10)
     pry (0.9.12.2)
       coderay (~> 1.0.5)
       method_source (~> 0.8)
@@ -28,10 +35,11 @@ GEM
       diff-lcs (>= 1.1.3, < 2.0)
     rspec-mocks (2.14.3)
     slop (3.4.6)
+    uuidtools (2.1.4)
     zk (1.9.2)
       logging (~> 1.7.2)
       zookeeper (~> 1.4.0)
-    zookeeper (1.4.6)
+    zookeeper (1.4.7)
 
 PLATFORMS
   ruby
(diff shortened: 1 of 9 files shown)


jolynch:
@igor47, @bobrik, @jamessharp  I'd like to revisit this because I sorta don't like adding the dynamic complexity directly to synapse/nerve. We use a cron job that reconfigures nerve/synapse every N seconds and I really like that approach because it keeps the design very unixy and easy to reason about IMO. At least for us it has made extending and debugging SmartStack really easy and I sorta want to keep it that way.

The only downside that I'm aware of is that synapse decides to restart haproxy the first time it starts and sometimes the watchers don't find all the backends immediately (why not sure). I think we can fix this using the state_file work that @jnb has done. Can anyone think of any other reasons that dynamically reacting to which watchers should exist should be synapse's responsibility and not an external component? 

If not then I'll close this PR and we'll continue down the state_file path.

brndnmtthws:
Feel free to close it. My 2 cents.

igor47:
agreed; closing.

PR #37, opened 2014-01-17
Add service watchers method

bakins:
This allows a user to add additional service watchers without hacking core.


jolynch:
I believe this is obsoleted by #132

PR #49, opened 2014-03-14
Add support for pushing data to udp services.

nherzing:
Because HAProxy does not support udp, an alternative is required. This commit
adds an optional "udp_forwarding" section to the service level configuration.
The only configuration is a port, over which udp data will be forwarded to one
of the service's host:port combinations. Both "udp_forwarding" and "haproxy" cannot be present.

Some example use cases of this would be forwarding data to an external logging service or external monitoring services.

This branch does not support incoming udp data but that felt like a rare use case to me.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,6 +3,7 @@ PATH
   specs:
     synapse (0.9.1)
       docker-api (~> 1.7.2)
+      eventmachine (~> 1.0.3)
       zk (~> 1.9.2)
 
 GEM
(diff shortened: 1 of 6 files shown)


jolynch:
@nherzing I believe that the [file_output](https://github.com/airbnb/synapse#file) mechanism is more elegant than this because it allows clients to do whatever they like with the registration information. Is there something in your use case that would not be satisfied by the file output?

For context we use the file output at Yelp to achieve this exact feature (services that talk UDP).

jolynch:
@nherzing I'm going to close this for now because I don't think that Synapse should be in the load balancing / traffic routing business itself (i.e. HAProxy does the load balancing, not Synapse).

I believe that the file output is sufficient for most udp based systems to load up themselves. If I'm missing something please do re-open this PR with an example and I can try to figure out another way to do this.

PR #53, opened 2014-03-26
Add FileWatcher watching a file containing server list.

ryotarai:
I added a new watcher, FileWatcher, which watches a server-list file. Because this watcher uses Listen gem, changes of the file will be detected immediately with something like inotify.

### Why I implemented this
I wrote this watcher for using with Serf. Serf can get join event and leave event. If a event handler of Serf adds (or removes) a server from the file, synapse will adds (or removes) the backend server to haproxy.cfg.
Of course, this works well without Serf too.


jolynch:
@ryotarai Can you rebase this against latest master and I can review if you're still interested in this watcher.

ryotarai:
Thank you for mention. Since I'm no longer interested in this feature, I close this PR

PR #58, opened 2014-04-13
Etcd service watcher

bobtfish:
And the etcd watcher part to go with the reporter :)

### DIFF
--- a/lib/synapse/service_watcher/etcd.rb
+++ b/lib/synapse/service_watcher/etcd.rb
@@ -9,14 +9,14 @@ module Synapse
     def start
       @etcd_hosts = @discovery['hosts'].shuffle
 
-      log.info "synapse: starting etcd watcher #{@name} @ host: #{@discovery['host']}, path: #{@discovery['path']}"
+      log.info "synapse: starting etcd watcher #{@name} @ hosts: #{@discovery['hosts']}, path: #{@discovery['path']}"
       @should_exit = false
 
       @etcd_hosts.each do |h|
         host, port = h.split(':')
         port = port || 4003
         @etcd = ::Etcd.client(:host => host, :port => port)
-        
+
         connected =
           begin
             @etcd.leader
@@ -52,10 +52,8 @@ module Synapse
     def validate_discovery_opts
       raise ArgumentError, "invalid discovery method #{@discovery['method']}" \
         unless @discovery['method'] == 'etcd'
-      raise ArgumentError, "missing or invalid etcd host for service #{@name}" \
-        unless @discovery['host']
-      raise ArgumentError, "missing or invalid etcd port for service #{@name}" \
-        unless @discovery['port']
+      raise ArgumentError, "missing or invalid etcd hosts for service #{@name}" \
+        unless @discovery['hosts']
       raise ArgumentError, "invalid etcd path for service #{@name}" \
         unless @discovery['path']
     end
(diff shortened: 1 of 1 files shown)


Zolmeister:
status?

SEJeff:
Well this has merge conflicts, so at a minimum needs to be rebased before it is merged.

WooDzu:
Any update on this?

tcolgate:
I've dropped my version of this PR. We've migrated to consul.

On 19 February 2015 at 08:35, Piotr Gasiorowski <notifications@github.com>
wrote:

> Any update on this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/airbnb/synapse/pull/58#issuecomment-75014877>.
>



-- 
Tristan Colgate-McFarlane
----
  "You can get all your daily vitamins from 52 pints of guiness, and a
glass of milk"

clizzin:
Sorry folks, the project maintainer @igor47 is really busy with other projects for Airbnb these days.  We're doing our best to find time to respond to pull requests, and we're sorry contributors are waiting so long. Please bear with us in the meantime.

apennebaker-ni:
+1

jolynch:
@bobtfish can you get a clean diff together and I can review/merge it?

bobtfish:
@jolynch - how do you feel about this? Pulled in #86 also :)

lib/synapse/service_watcher/etcd.rb
```diff
@@ -0,0 +1,166 @@
+require "synapse/service_watcher/base"
+
+require 'etcd'
+
+module Synapse
```
jolynch: This should be class Synapse::ServiceWatcher

Gemfile.lock
```diff
@@ -21,6 +22,8 @@ GEM
       archive-tar-minitar
       excon (>= 0.28)
       json
+    etcd (0.2.4)
```
jolynch: So while testing this I came across quite a pickle, which is that the etcd ruby gem < 0.3.0 cannot handle etcd 2.0+ (it errors all over the place regarding 404 errors)

https://github.com/ranjib/etcd-ruby/pull/51 has been merged and my local testing indicates that version 0.3.0 seems to fix things.

@bobtfish do you think we should depend on an old etcd or the new one? I sorta prefer the new one but I'm not sure what the community progress on moving to etcd 2.0 is like?

lib/synapse/service_watcher/etcd.rb
```diff
@@ -0,0 +1,166 @@
...
+          log.info "synapse: discovered #{new_backends.length} backends for service #{@name}"
+          false
+        end
+      end
+    end
+
+    def watch
+      while !@should_exit
+        begin
+          @etcd.watch(@discovery['path'], :timeout => 60, :recursive => true)
```
jolynch: Alright so while testing I came across this gem. If my understanding is correct etcd makes us not only do our own heartbeats from nerve (hella writes), but we also end up basically constantly polling in this function because there is no capability to filter watch events?

Basically without https://github.com/coreos/etcd/issues/633 or https://github.com/coreos/etcd/issues/174 closed SmartStack scalability on etcd will be limited to ... not very much.

If that understanding is correct can we implement get children watches internally, where we do a lightweight "see if the list of children changed" operation and if that is different we actually go through loading all the data (aka run discover). This way we're at least not constantly pulling all of the etcd state?

Gemfile.lock
jolynch: Now that we recommend installing via gem, can we just do ~> 0.2 ?

jolynch:
@bobtfish ping?

jolynch:
Looks like the v3 etcd API may have solved some of the fundamental issues. I'll take another look at this in the next few days.

philips:
@jolynch if you have any problems/questions feel free to ping https://groups.google.com/forum/#!forum/etcd-dev or @heyitsanthony

PR #72, opened 2014-06-11
Added global zookeeper for service discovery

smarth-madan:
Added global zookeeper section in the config file. If each service has zookeeper hosts present, a new connection will be opened for that service. If the service discovery is still through zookeeper and the hosts are not present, global zookeeper hosts will be used to limit the no. of open connections to the same zookeeper instance.


jolynch:
I believe this is obsoleted by the pooling code in #130

PR #79, opened 2014-07-10
Active version watcher

qasimkhan:
To avoid the collision of different releases (as by default all nodes share the same path) in immutable infrastructure we need some sort of identifier which tells about the active version so we can switch between different releases.

This watcher is build on top of ZK watcher. It will watch the active version along with the nodes and as soon as a version changes it will update the synapse config with latest version and reload the haproxy. 
e.g. version 1 nodes are resgitered in the following path using nerve `/services/service1/v1` and the active version path `/version/service1` is set to v1. All version 2 nodes will go to `/service/service1/v2` and once all the nodes are ready we can switch version to v2 on `/version/service1`

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,21 @@ The watcher assumes that each node under `path` represents a service server.
 Synapse attempts to decode the data in each of these nodes using JSON and also using Thrift under the standard Twitter service encoding.
 We assume that the data contains a hostname and a port for service servers.
 
+##### Version #####
+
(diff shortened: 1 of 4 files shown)


jolynch:
@qasimkhan I'm not sure I see why this needs a separate watcher. Can't you use a shared frontend with multiple backends and use an ACL on anything at all in the Requests to switch between them?

jolynch:
Closing due to inactivity, please do re-open if you still want this :-)

PR #82, opened 2014-07-16
Allow service-specific bind_addresses

vulpine:
While one can specify bind_address in the haproxy section of the config, this applies to all frontends globally. I would like to be able to override this per frontend, by specifying it with the port in the haproxy section of my service thus:

```
      "haproxy": {
        "bind_address": "1.2.3.4",
        "port": "80",
        ....
```


jolynch:
Got merged with the Yelp fork over in #130

jolynch:
@vulpine Whoops I was mistaken to close this. I like this, can you rebase and add a test?

hashbrowncipher:
I believe this happened in #156.

jolynch:
Indeed it did

PR #83, opened 2014-07-20
slaves_as_backup: mark the slave nodes as backup nodes in haproxy

Dareen:
This way the slave nodes will be used as backup nodes if the master node fails while zookeeper is unreachable by synapse.


jolynch:
I believe that #140 provides this functionality, although it doesn't do the leader detection. I'm going to close this but if you can't achieve what you want with `haproxy_server_options` please do let me know :-)

PR #84, opened 2014-08-05
Make AWS region, access key, and secret key settings optional.

liamstewart:
The AWS SDK is smart enough to figure out the region, access key, and secret
key if the user has not specified them. It looks for the AWS_REGION environment
variable itself. It also uses a credential provider that attempts to locate
credentials from 1) AWS.config, 2) environment variables, 3) the EC2 metadata
service. Using the built-in configuration functionality of the SDK is required
for instances that get credentials based on IAM roles and have no access keys
on the instance.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    synapse (0.11.1)
+    strava-synapse (0.11.3)
       aws-sdk (~> 1.39)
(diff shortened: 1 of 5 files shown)


jolynch:
@liamstewart I believe that when I merged #126 I fixed this. I'm going to close this PR but if master does not work for you (we'll be cutting a release soon), please re-open an issue or PR and I can take a look.

PR #87, opened 2014-08-17
Marathon Watcher

deverton:
This adds a service watcher that uses data from the Marathon REST API V2 to locate services. It's based on a combination of the Docker and EC2 watchers.


jolynch:
Merged #101 which should cover this functionality

PR #92, opened 2014-09-04
Filter out containers with unmapped ports

cap10morgan:
This bit us when running one-off containers on our docker hosts (to run scripts, debug problems, etc.). Synapse would pick up the container because the image name and the private port matched, but the missing public port would cause it to generate an invalid haproxy config file and crash.

This fixes that by adding a simple check that the public port is present when deciding which containers to consider adding as backends.

It only does this for new-style port mappings because I can't find any docs on what the old-style port mappings looked like in this situation, nor is it clear that you could even prevent an exposed port from being mapped back when that version of the remote API was in use.

### DIFF
--- a/lib/synapse/service_watcher/docker.rb
+++ b/lib/synapse/service_watcher/docker.rb
@@ -80,10 +80,11 @@ module Synapse
         cnts.each do |cnt|
           cnt['Ports'] = rewrite_container_ports cnt['Ports']
         end
-        # Discover containers that match the image/port we're interested in
+        # Discover containers that match the image/port we're interested in and have the port mapped to the host
         cnts = cnts.find_all do |cnt|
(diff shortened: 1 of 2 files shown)


cap10morgan:
Any chance of getting this merged?

jolynch:
@cap10morgan Looks reasonable to me, thanks for the PR

cap10morgan:
@jolynch Great, thanks! Can this go out in a new release sometime soon?

jolynch:
@cap10morgan unfortunately I don't have commit access to the gem on https://rubygems.org/gems/synapse

We've always just built from source to work around this. @igor47 do you think you can cut a new release?

igor47:
i pushed v0.12.2 to rubygems

PR #98, opened 2014-11-10
AWS credentials are not required when 'use_iam_profile' is specified

dylanmei:
Similar to #84 except that
- `aws_region` continues to be validated (the aws-sdk will not auto-magically fill this in.)
- `use_iam_profile` is an explicit configuration option

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,13 @@ be used in preference to the `AWS_` environment variables.
 * `aws_secret_access_key`: AWS secret key or set `AWS_SECRET_ACCESS_KEY` in the environment.
 * `aws_region`: AWS region (i.e. `us-east-1`) or set `AWS_REGION` in the environment.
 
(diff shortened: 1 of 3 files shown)


liammac:
Can someone please merge this pull request? Having to expose the secret is non optimal

igor47:
i don't think we need `use_iam_profile` -- we can just avoid requiring the creds if they're not specified. that's the typical behavior for AWS tools that can automatically use IAM roles.

dylanmei:
Great, then we just need to stop demanding that they are defined.
https://github.com/airbnb/synapse/blob/master/lib/synapse/service_watcher/ec2tag.rb#L49-L53

jolynch:
@dylanmei I believe that #126 makes it so we still insisting they are defined. Can you test out if master works for you and if so close this PR?

dylanmei:
:+1:

PR #101, opened 2014-11-27
Add marathon watcher

tdooner:
I wrote this before I noticed #87, so in case you'd like a version that doesn't add a runtime dependency on `marathon_client` then this is your PR. Otherwise these are pretty much identical PRs and I'll close this out if you merge the other one.

### DIFF
--- a/lib/synapse/service_watcher.rb
+++ b/lib/synapse/service_watcher.rb
@@ -4,6 +4,7 @@ require "synapse/service_watcher/ec2tag"
 require "synapse/service_watcher/dns"
 require "synapse/service_watcher/docker"
 require "synapse/service_watcher/zookeeper_dns"
+require "synapse/service_watcher/marathon"
 
 module Synapse
   class ServiceWatcher
@@ -15,6 +16,7 @@ module Synapse
       'dns' => DnsWatcher,
       'docker' => DockerWatcher,
       'zookeeper_dns' => ZookeeperDnsWatcher,
+      'marathon' => MarathonWatcher,
     }
 
     # the method which actually dispatches watcher creation requests
--- a/lib/synapse/service_watcher/marathon.rb
+++ b/lib/synapse/service_watcher/marathon.rb
@@ -1,0 +1,82 @@
+require "synapse/service_watcher/base"
+require 'json'
+require 'resolv'
+
+module Synapse
+  class MarathonWatcher < BaseWatcher
+    def start
+      @check_interval = @discovery['check_interval'] || 10.0
+      @watcher = Thread.new { watch }
+
+      @marathon_api = URI.join(@discovery['marathon_api_url'], "/v2/apps/#{@discovery['application_name']}/tasks")
+      @connection = Net::HTTP.new(@marathon_api.host, @marathon_api.port)
+      @connection.start
+    end
+
+    def stop
+      @connection.finish
+    rescue
+      # pass
+    end
+
+  private
+
+    def validate_discovery_opts
+      required_opts = %w[marathon_api_url application_name]
+
+      required_opts.each do |opt|
+        if @discovery.fetch(opt, '').empty?
+          raise ArgumentError,
+            "a value for services.#{@name}.discovery.#{opt} must be specified"
+        end
+      end
+    end
+
+    def watch
+
+      until @should_exit
+        retry_count = 0
+        start = Time.now
+
+        begin
+          response = @connection.request Net::HTTP::Get.new(@marathon_api.request_uri)
+
+          tasks = JSON.parse(response.body).fetch('tasks', [])
+          backends = tasks.map do |task|
+            {
+              'name' => task['host'],
+              'host' => task['host'],
+              'port' => task['ports'].first,
+            }
+          end
+
+          if backends.empty?
+            log.warn "synapse: no backends discovered for #{@discovery['application_name']}"
+          else
+            previous_backends = @backends
+            set_backends(backends)
+            new_backends = @backends
+
+            unless previous_backends == new_backends
+              log.info "synapse: found #{backends.length} backends for #{@discovery['application_name']}"
+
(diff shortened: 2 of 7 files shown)


lib/synapse/service_watcher.rb
```diff
@@ -4,6 +4,7 @@
 require "synapse/service_watcher/dns"
 require "synapse/service_watcher/docker"
 require "synapse/service_watcher/zookeeper_dns"
+require "synapse/service_watcher/marathon"
```
jolynch: Can you rebase to pick up the auto-loading watcher code? See https://github.com/airbnb/synapse/blob/master/lib/synapse/service_watcher/README.md for the new specification

lib/synapse/service_watcher/marathon.rb
```diff
@@ -0,0 +1,105 @@
...
+      required_opts.each do |opt|
+        if @discovery.fetch(opt, '').empty?
+          raise ArgumentError,
+            "a value for services.#{@name}.discovery.#{opt} must be specified"
+        end
+      end
+    end
+
+    def attempt_marathon_connection
+      @marathon_api = URI.join(@discovery['marathon_api_url'], "/v2/apps/#{@discovery['application_name']}/tasks")
```
jolynch: Can we make the uri configurable? I'm not sure how often marathon changes it's api but I imagine v2 will only be valid for a little while longer.

lib/synapse/service_watcher/marathon.rb
```diff
@@ -0,0 +1,105 @@
...
+            set_backends(backends)
+            new_backends = @backends
+
+            unless previous_backends == new_backends
+              log.info "synapse: found #{backends.length} backends for #{@discovery['application_name']}"
+
+              reconfigure!
+            end
+          end
+        rescue EOFError
```
jolynch: Do we need to set @connection to nil here to ensure we attempt to connect back to marathon?

lib/synapse/service_watcher/marathon.rb
```diff
@@ -0,0 +1,105 @@
+require "synapse/service_watcher/base"
+require 'json'
+require 'resolv'
+
+module Synapse
+  class MarathonWatcher < BaseWatcher
+    def start
+      @check_interval = @discovery['check_interval'] || 10.0
+      @connection = nil
+      @watcher = Thread.new { watch }
```
jolynch: Can we sleep for the first watch for a random amount of time between 0 and the check_interval? I feel like this could easily thunder your marathon api?

lib/synapse/service_watcher/marathon.rb
```diff
@@ -0,0 +1,105 @@
...
+            log.warn "synapse: no backends discovered for #{@discovery['application_name']}"
+          else
+            previous_backends = @backends
+            set_backends(backends)
+            new_backends = @backends
+
+            unless previous_backends == new_backends
+              log.info "synapse: found #{backends.length} backends for #{@discovery['application_name']}"
+
+              reconfigure!
```
jolynch: Don't call reconfigure again, set_backends takes care of this for you (taking care of things like de-duplicating host/port pairs and such).

jolynch:
@tdooner Awesome, I think with a little bit of cleanup this is good for merge.

Since I'm not too worried about changes in the marathon API (like I am with etcd, docker, etc ...) that might make this direct request approach problematic long term, I think I prefer this to one that adds a dependency on marathon_client. What are your thoughts having run marathon in production for a while? Does it seem like they are good about keeping backwards compatibility?

lib/synapse/service_watcher/marathon.rb
```diff
@@ -0,0 +1,105 @@
...
+          req = Net::HTTP::Get.new(@marathon_api.request_uri)
+          req['Accept'] = 'application/json'
+          response = @connection.request(req)
+
+          tasks = JSON.parse(response.body).fetch('tasks', [])
+          backends = tasks.keep_if { |task| task['startedAt'] }.map do |task|
+            {
+              'name' => task['host'],
+              'host' => task['host'],
+              'port' => task['ports'].first,
```
jolynch: Can you do a port_index like the other PR? I like that flexibility.

lib/synapse/service_watcher/marathon.rb
tdooner: Idk? It seems pretty stable to me, and I think even if they were to make a breaking change to it they might still maintain the old URL. Nonetheless, I will make it configurable so I don't have to eat my words.

lib/synapse/service_watcher/marathon.rb
tdooner: Yeah, I think so.

lib/synapse/service_watcher/marathon.rb
tdooner: Done

lib/synapse/service_watcher/marathon.rb
tdooner: Done. (7d512887 moved that logic into `ServiceWatcher::Base`, after my first version of the PR)

lib/synapse/service_watcher/marathon.rb
tdooner: Good idea, we actually needed this too.

tdooner:
All done!

My thoughts on having run this in production: it is a lot of requests to the Marathon API, since it scales up by the product of the number of machines and the number of services you have. For small deployments and relatively-few backend tasks, Marathon will be able to take the heat. At Yelp or Airbnb scale? This won't work.

At the time of writing this initially, only the Marathon eventSubscriptions existed -- this was a pretty clunky way to request Marathon POST event updates as they happen. This isn't a good fit for Synapse. But since then, Marathon has introduced the `/events` endpoint, which streams these events via HTTP. Supporting this endpoint seems like a great next step for this service watcher.

PR #102, opened 2014-12-02
virtual/indirect backends in zookeeper

ajf8:
Hi :)

I'm looking at using smartstack to replace our current mess of apache, mod_jk, and mod_rewrite rules which we use for some fairly complex SOA routing. We think smartstack is great and I've been able to get a lot working, integrated nicely with puppet. Thanks for sharing it.

Anyhow, one of the things we do a lot of is "blue/green" deployments. We test the dark deployment by routing to it using a special query parameter or cookie, and once it's OK our deployment scripts update a mod_rewrite map file which causes traffic to start being routed to it. Maybe we will move beyond the concept of "blue" or "green" to versions in the future.

I couldn't figure out a way to do this easily with nerve/synapse (would be interested if anyone else does this). So in the zookeeper backend I've added an optional level of indirection which you can use with the "indirect_path" discovery parameter. It sets up a watch on this node, as well as the one it refers to.

I thought I'd share it, just in case anyone else likes the idea.

So nerve registers into 

/nerve/services/marketlist/blue and /nerve/services/marketlist/green

synapse has this configured for the dark service:

"indirect_path":"/nerve/services/marketlist/dark",
...
      "shared_frontend":[
         "acl url_marketlist path_beg /marketlist",
         "acl url_marketlist_dark url_sub marketlist_dark",
         "use_backend marketlist_dark if url_marketlist url_marketlist_dark"
      ],

and this for live:

```
  "indirect_path":"/nerve/services/marketlist/live",
```

...
      "shared_frontend":[
         "acl url_marketlist path_beg /marketlist",
         "use_backend marketlist_live if url_marketlist"
      ],

so in zookeeper

set /nerve/services/marketlist/dark /nerve/services/marketlist/blue
set /nerve/services/marketlist/live /nerve/services/marketlist/green

And now we can switch these round easily.

### DIFF
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -12,7 +12,12 @@ module Synapse
       @watcher = nil
       @zk = nil
 
-      log.info "synapse: starting ZK watcher #{@name} @ hosts: #{@zk_hosts}, path: #{@discovery['path']}"
+      if @discovery['indirect_path']
+        log.info "synapse: starting ZK watcher #{@name} @ hosts: #{@zk_hosts}, indirect_path: #{@discovery['indirect_path']}"
+      else
+        log.info "synapse: starting ZK watcher #{@name} @ hosts: #{@zk_hosts}, path: #{@discovery['path']}"
+      end
+
       zk_connect
     end
 
@@ -33,7 +38,7 @@ module Synapse
       raise ArgumentError, "missing or invalid zookeeper host for service #{@name}" \
         unless @discovery['hosts']
       raise ArgumentError, "invalid zookeeper path for service #{@name}" \
-        unless @discovery['path']
+        unless @discovery['path'] or @discovery['indirect_path']
     end
 
     # helper method that ensures that the discovery path exists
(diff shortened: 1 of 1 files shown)


jolynch:
@ajf8 Thank you for the pull request, but I'm a bit confused for why synapse needs to watch two paths on a service to allow for this. It seems to me that this is workable with just frontend/backend sections and ACLs in HAProxy. I'm thinking:

1. Setup a backend "myservice" that has a port (so it'll get a frontend and a backend)
2. Setup a backend "myservice_dark" that does not have a port (so it'll just get a backend)
3. Use the backend_name from https://github.com/airbnb/synapse#the-haproxy-section so that you can reference 1 and 2 from the frontend in 1.
4. Setup acls as per your suggestion (basically rout to myservice_dark if a certain path or cookie is set, otherwise service)
5. Give nerve the smarts for where to register what code, anything that is live you can register for myservice, and anything that is dark goes in myservice_dark.

Would that work?

ajf8:
Hi,

1-4 all sounds fine, and 5 sounds like something that could work, but I thought it made more sense to keep an indirection in zookeeper, as a central place to update in one simple step from a deployment script or some other CI infrastructure. The aim in a bluegreen deployment is to do the switchover as transparently as possible to the user, and only after you've tested the "dark" side is working correctly. Doing it on the nerve side would be reconfiguring pools before testing, and that would require far more complex orchestration in a cluster of backends than a simple update in zookeeper. Also more difficult to do it in an atomic way I would guess, with the potential to have your blues/greens mixed up if something went wrong.

But... I did this work 9 months ago at a previous employer, we did a pretty big successful smartstack rollout, and then I left the company shortly after back in May. So I'm not working with smartstack at all anymore, and if this feature isn't interesting to you then please feel free to close this PR :)

PR #104, opened 2014-12-05
watcher for Apache Aurora zookeeper announcements

benley:
Aurora zookeeper serversets are comparable to what Nerve does, but Aurora uses a different json structure in the zookeeper nodes, and it has provisions for more than one advertised port for a service.

[Review on Reviewable](https://reviewable.io/reviews/airbnb/synapse/104)

### DIFF
--- a/lib/synapse/service_watcher/zookeeper_aurora.rb
+++ b/lib/synapse/service_watcher/zookeeper_aurora.rb
@@ -1,0 +1,76 @@
+# encoding: utf-8
+#
+# Variant of the Zookeeper service watcher that works with Apache Aurora
+# service announcements.
+#
+# Parameters:
+#   hosts: list of zookeeper hosts to query (List of Strings, required)
+#   path: "/path/to/serverset/in/zookeeper" (String, required)
+#   port_name: Named service endpoint (String, optional)
+#
+# If port_name is omitted, uses the default serviceEndpoint port.
+
+# zk node data looks like this:
+#
+# {
+#   "additionalEndpoints": {
+#     "aurora": {
+#       "host": "somehostname",
+#       "port": 31943
+#     },
+#     "http": {
+#       "host": "somehostname",
+#       "port": 31943
+#     },
+#     "otherport": {
+#       "host": "somehostname",
+#       "port": 31944
+#     }
+#   },
+#   "serviceEndpoint": {
+#     "host": "somehostname",
+#     "port": 31943
+#   },
+#   "shard": 0,
+#   "status": "ALIVE"
+# }
+#
+
+require 'synapse/service_watcher/zookeeper'
+
+module Synapse
+  # Watcher for Zookeeper announcements from Apache Aurora
+  class ZookeeperAuroraWatcher < Synapse::ZookeeperWatcher
+    def validate_discovery_opts
+      @discovery['method'] == 'zookeeper_aurora' ||
+        fail(ArgumentError,
+             "Invalid discovery method: #{@discovery['method']}")
+      @discovery['hosts'] ||
+        fail(ArgumentError,
+             "Missing or invalid zookeeper host for service #{@name}")
+      @discovery['path'] ||
+        fail(ArgumentError, "Invalid zookeeper path for service #{@name}")
+    end
+
+    def deserialize_service_instance(data)
+      log.debug 'Deserializing process data'
+      decoded = JSON.parse(data)
+
+      name = decoded['shard'].to_s ||
+        fail("Instance JSON data missing 'shard' key")
+
+      hostport = if @discovery['port_name']
+                   decoded['additionalEndpoints'][@discovery['port_name']] ||
+                     fail("Endpoint '#{@discovery['port_name']}' not found " \
+                          'in instance JSON data')
+                 else
+                   decoded['serviceEndpoint']
+                 end
+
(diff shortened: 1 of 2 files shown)


larryweya:
@benley  I like your implementation, allowing a custom port_name, since the rest is exactly the same, I'll merge yours into mine.

StephanErb:
Anything missing to get this one merged?

StephanErb:
@jolynch you have just merged a Marathon watcher (#101) for the Marathon Mesos Framework. This one implements the same for the Apache Aurora Mesos framework. Any change to get this one here merged as well? The code is rock solid; we are already using it in production.

benley:
I can update this to fit the new plugin model pretty easily if there's actually a chance of getting it merged.

jolynch:
Yes I'm trying to work my way through the pull requests. If you pull/rebase I can review this one next. I think Igor is handling review of the exhibitor watcher, and we have a few more to review.

StephanErb:
@benley two minor remarks:

* This watcher does not only support Aurora but also others, such as finagle. Renaming it to serverset watcher would make that more apparent.
* The pull request is missing a short description in the REAMDE https://github.com/airbnb/synapse#service-discovery 

Really looking forward to seeing this merged. Thanks for your effort :+1:

benley:
I agree about renaming it to serverset watcher; I'll get the PR updated soon.

lib/synapse/service_watcher/zookeeper_aurora.rb
```diff
@@ -0,0 +1,76 @@
...
+#       "host": "somehostname",
+#       "port": 31944
+#     }
+#   },
+#   "serviceEndpoint": {
+#     "host": "somehostname",
+#     "port": 31943
+#   },
+#   "shard": 0,
+#   "status": "ALIVE"
```
jolynch: Should we take into account the ALIVE status?

lib/synapse/service_watcher/zookeeper_aurora.rb
```diff
@@ -0,0 +1,76 @@
+# encoding: utf-8
+#
+# Variant of the Zookeeper service watcher that works with Apache Aurora
+# service announcements.
+#
+# Parameters:
+#   hosts: list of zookeeper hosts to query (List of Strings, required)
+#   path: "/path/to/serverset/in/zookeeper" (String, required)
+#   port_name: Named service endpoint (String, optional)
```
jolynch: Can we call this 'endpoint_name'? I'm a little confused by 'port_name' what it actually does

lib/synapse/service_watcher/zookeeper_aurora.rb
```diff
@@ -0,0 +1,76 @@
...
+      @discovery['path'] ||
+        fail(ArgumentError, "Invalid zookeeper path for service #{@name}")
+    end
+
+    def deserialize_service_instance(data)
+      log.debug 'Deserializing process data'
+      decoded = JSON.parse(data)
+
+      name = decoded['shard'].to_s ||
+        fail("Instance JSON data missing 'shard' key")
```
jolynch: can you raise a ValueError to be consistent with the zk watcher?

lib/synapse/service_watcher/zookeeper_aurora.rb
```diff
@@ -0,0 +1,76 @@
...
+                     fail("Endpoint '#{@discovery['port_name']}' not found " \
+                          'in instance JSON data')
+                 else
+                   decoded['serviceEndpoint']
+                 end
+
+      host = hostport['host'] || fail("Instance JSON data missing 'host' key")
+      port = hostport['port'] || fail("Instance JSON data missing 'port' key")
+
+      [host, port, name]
```
jolynch: The interface for this function has changed in master, and I'll make a PR to support haproxy_server_options so we only have to update this once. The spec for service registration data is now roughly defined by https://github.com/airbnb/synapse/blob/master/lib/synapse/service_watcher/README.md#backend-interface

Just a heads up.

lib/synapse/service_watcher/zookeeper_aurora.rb
StephanErb: Looking at Aurora, I am not sure if the status is ever toggled. Aurora seems to simply remove the znode when the registered server disconnects. I'd say we leave it out for now.

lib/synapse/service_watcher/zookeeper_aurora.rb
StephanErb: Serversets allow you to register multiple ports for your service. For example, you can say that your service has an `http`, an `https`and a `admin` port. This option can be used to specify which port shall actually be used by synapse.

jolynch:
See #144 for the current deserialize interface

lib/synapse/service_watcher/zookeeper_aurora.rb
jolynch: I see, ok I'll leave the naming up to you guys :-)

benley:
I'm planning on working on this PR tomorrow afternoon-ish, fwiw

jolynch:
I believe that #153 should support what this PR requests via the new `decode` option to the `ZookeeperWatcher`. The README has details for how to configure it.

If that doesn't work please open and issue/PR and we can work on it :-)

benley:
Thanks for finishing this work, @jolynch!  I got buried with other projects and totally lost track of it.

PR #106, opened 2014-12-11
Add 'use_previous_backends' option.

jnb:
* This option defaults to true:  if there are no default servers and a watcher
  reports no backends, then use the previous backends that we already know
  about.
* Factor out code that updates `@backends` into the watcher base class.

This code is running in production with no issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,10 @@ If you do not list any default servers, no proxy will be created.  The
 `default_servers` will also be used in addition to discovered servers if the
 `keep_default_servers` option is set.
 
+If you do not list any `default_servers`, and all backends for a service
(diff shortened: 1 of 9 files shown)


igor47:
this is great; i've been wanting to do a refactor like this for awhile.

@jnb we talked about you becoming a maintainer of this repo. are you still interested? i have a few days this week to focus squarely on synapse, so i'm going to work on clearing out the current backlog and adding more unit testing. i've already done this for airbnb/nerve last week.

jolynch:
Note, if you accept #115, that supersedes this change.

igor47:
cool, we'll just work with #115

PR #108, opened 2015-01-23
Update configs

ajohnstone:


config/hostheader_test.json
```diff
@@ -8,8 +8,16 @@
         "method": "dns",
         "nameserver": "127.0.0.1",
         "servers": [
-          "0.www.example.com",
-          "1.www.example.com"
+          {
+            "name": "example-service",
+            "port": 80,
```
jolynch: Do you actually want name or port? I'm not terribly well versed in the dns watcher but it seems that it only uses the server's ``host`` key?

PR #111, opened 2015-02-06
added Dockerfile

Zolmeister:


Zolmeister:
Forked, let me know if you still want the Dockerfile though

Issue #94, opened 2014-10-09
Synapse fails to restart after zookeeper node disappears


Jaykah:
I am using the Zookeeper watcher to get a list of available nodes - which it successfully does.

However, after a node fails, it does not get removed from haproxy.

Haproxy monitoring page shows those nodes as down, Zookeeper no longer has them as they are ephemeral, but the reconfiguration does not take place.

If I restart Synapse, however, everything goes back to normal.

```
I, [2014-10-09T22:17:20.844370 #12]  INFO -- Synapse::ZookeeperWatcher: synapse: discovering backends for service proddb
I, [2014-10-09T22:17:20.861672 #12]  INFO -- Synapse::ZookeeperWatcher: synapse: discovered 6 backends for service proddb
[WARNING] 281/221756 (157) : Server proddb/100.0.10.10:3306_100.0.10.10 is DOWN, reason: Layer4 connection problem, info: "Connection refused", check duration: 1ms. 6 active and 0 backup servers left. 0 sessions active, 0 requeued, 0 remaining in queue.
[WARNING] 281/221756 (157) : Server proddb/100.0.10.10:3306_100.0.10.10 is DOWN, reason: Layer4 connection problem, info: "Connection refused", check duration: 2ms. 5 active and 0 backup servers left. 0 sessions active, 0 requeued, 0 remaining in queue.
I, [2014-10-09T22:18:15.682057 #12]  INFO -- Synapse::ZookeeperWatcher: synapse: discovering backends for service proddb
I, [2014-10-09T22:18:15.690843 #12]  INFO -- Synapse::ZookeeperWatcher: synapse: discovered 5 backends for service proddb
```

Note: Synapse is deployed in Docker and managed by supervisord

hankya:
is this fixed? anyone?

jolynch:
@hankya We haven't seen the behavior referenced in this PR but we don't use the supervisord. Have you been seeing the same issue and if so what's your setup?

I did see something that I thought was like this bug when ZK was overloaded, but in that case the issue was just a large lag in event delivery. Our connection pooling fix alleviated those issues (plus a bug fix to the pooling code). While investigating the pooling bug though I setup a pretty thorough test setup where I was removing and adding nodes while messing with the network via netem and I still wasnt able to repro with Ruby 1.9 and running Synapse directly. I could try to modify that setup to get a repro of this issue if it's still impacting people.

Jaykah:
[...]

PR #115, opened 2015-03-12
Allow registrations to be manifested on the file system and add 'use_previous_backends' option

jolynch:
This pull request pulls in the Yelp fork of synapse. It accomplishes:

1. Merges in https://github.com/airbnb/synapse/pull/106, which provides the use_previous_backends option.
2. Provide the "file_output" option in the configuration, which will cause synapse to manifest registrations on the filesystem. This is really useful for services that do not or can not use haproxy for routing requests (e.g. cassandra, udp services, etc ...). By default this is disabled and only haproxy configuration is written out.

Both of these changes are functioning in production at Yelp.


jolynch:
To give some context on the allredisp option:

http://marc.info/?l=haproxy&m=142740715826651&w=2

To give some context on the deduplication, we currently restart nerve gracefully (i.e. not de-registering a whole box) by starting two nerves and relying on synapse to deduplicate.

Jaykah:
Hi Joseph,

First of all, thanks for the PR! 

Secondly, I have a question on deduplication. How does it work with process managers that control forking? 

Precisely, I've had issues with running synapse via Supervisord (#94) where it would not restart when a change in backends is detected, unless I manually specify to run 4 processes of Synapse in Supervisord.

jolynch:
My deduplication comment was not the clearest. The deduping logic is just referring to synapse ignoring duplicated service backends, e.g. (host1, 1234), (host1, 1234), (host2, 1234) => (host1, 1234), (host2, 1234).

I would want to run some tests since supervisord is not the setup we use (we use a cron job that restarts synapse whenever our synapse config file changes), but the deduplication is implemented by synapse filtering out duplicate backends by the tuple of (host, port, name?), so I imagine that all of the forked synapse instances would deduplicate in the same way. To be honest I am not sure without digging into #94 more.

lib/synapse/file_output.rb
```diff
@@ -0,0 +1,53 @@
...
+require 'fileutils'
+require 'tempfile'
+
+module Synapse
+  class FileOutput
+    include Logging
+    attr_reader :opts, :name
+
+    def initialize(opts)
+      super()
```
igor47: not sure what this is calling out to, this class does not inherit from anything

lib/synapse/service_watcher/base.rb
```diff
@@ -95,13 +100,44 @@ def validate_discovery_opts
     end
 
     def set_backends(new_backends)
-      if @keep_default_servers
-        @backends = @default_servers + new_backends
+      # Aggregate and deduplicate all potential backend service instances.
```
igor47: are you really concerned about duplicates in the default servers list? if that's the case, maybe we should de-dup `@default_servers`  when we set it

lib/synapse/service_watcher/base.rb
```diff
@@ -95,13 +100,44 @@ def validate_discovery_opts
...
+              " 'use_previous_backends' is disabled; dropping all backends"
+            @backends.clear
+          end
+        else
+          log.warn "synapse: no backends for service #{@name};" \
+            " using default servers: #{@default_servers.inspect}"
+          @backends = @default_servers
+        end
       else
+        log.info "synapse: discovered #{new_backends.length} backends for service #{@name}"
```
igor47: on an unrelated note, one thing i always thought might be necessary eventually is to specify the minimum number of backends to serve traffic.

suppose a service goes down. when functionality is restored, a single backend is naturally going to be the first to come up. that backend might then be pounded with all traffic which is normally meant for many multiple backends, and as a result might go down because of excess traffic. this would then happen to the second backend, etc.. etc.. 

it might be nice here to avoid setting any backends unless the number exceeds some limit which we know can handle the traffic. just a brain-dump, not even necessarily a todo.

lib/synapse/service_watcher/dns.rb
```diff
@@ -89,21 +89,7 @@ def configure_backends(servers)
...
-            " using default servers: #{@default_servers.inspect}"
-          @backends = @default_servers
-        end
-      else
-        log.info "synapse: discovered #{new_backends.length} backends for service #{@name}"
-        set_backends(new_backends)
-      end
-
-      reconfigure!
+      set_backends(new_backends)
```
igor47: great!

lib/synapse/service_watcher/base.rb
```diff
@@ -95,13 +100,44 @@ def validate_discovery_opts
...
+        [b['host'], b['port'], b.fetch('name', '')]
+      }
+
+      if new_backends.to_set == @backends.to_set
+        return false
+      end
+
+      if new_backends.empty?
+        if @default_servers.empty?
+          if @use_previous_backends
```
igor47: thanks for addressing this original "mis-feature"

lib/synapse/haproxy.rb
```diff
@@ -386,6 +388,7 @@ class Haproxy
         "option accept-invalid-http-request",
         "option accept-invalid-http-response",
         "option allbackups",
+        "option allredisp",
```
igor47: should we try to re-sync these lists of options with latest haproxy? i wonder what other options have been added since i put these lists in here...

lib/synapse/service_watcher/base.rb
```diff
@@ -95,13 +100,44 @@ def validate_discovery_opts
     end
 
     def set_backends(new_backends)
-      if @keep_default_servers
-        @backends = @default_servers + new_backends
+      # Aggregate and deduplicate all potential backend service instances.
+      new_backends = (new_backends + (@keep_default_servers ? @default_servers : [])).uniq {|b|
```
igor47: oh i see. i think this would read a little cleaner if it was not all on one line. i would say:
```ruby
new_backends += @default_servers if @keep_default_servers
new_backends.uniq!{|b|  [b['host'], b['port'], b.fetch('name', '')]}
```

igor47:
done with first pass, i left a couple of questions/comments. mostly, it looks great!

lib/synapse/file_output.rb
jolynch: Indeed it does not, I'll fix that up

lib/synapse/service_watcher/base.rb
jolynch: I'm concerned with duplicates in the set of backends we write out to HAProxy in general. Like, we had a few cases where duplicate backends got written out and then synapse would only down a single one when we failed healthchecks.

If we dedup the default_servers separately doesn't that mean that we might end up with duplicates between the defaults and new_backends and could end up with the bad situation I mention above?

lib/synapse/service_watcher/base.rb
jolynch: I agree this is a needed feature. I've been hacking on a "latency sensitive farm failover" method as suggested in http://blog.haproxy.com/2013/12/23/failover-and-worst-case-management-with-haproxy/ and determining how many backends are needed is core to that.

The idea is that you have multiple HAProxy backends powering a single frontend, with each backend having more nodes (e.g. because it encompasses multiple datacenters).

lib/synapse/haproxy.rb
jolynch: Heh, this isn't even the option that I ended up getting accepted. http://marc.info/?l=haproxy&m=143161966505891&w=2 basically added a parameter to "option redispatch" which allows this behavior. Until that is released in 1.6 do you mind if we keep this in here (we use is pretty heavily at Yelp)

jolynch:
Sorry that I've just been adding a ton more commits to this "de-fork" pull-request. We ended up hitting some serious reliability problems when we turned on some new multi-datacenter functionality and had to focus on fixing that up. I'll be getting this ready for merge ASAP.

jolynch:
We're going to take care of this in #130

PR #116, opened 2015-03-17
Update Rakefile

worldsouptester:


jolynch:
@worldsouptester What's wrong with the Rakefile?

PR #126, opened 2015-06-24
ec2tag watcher improvements

dcosson:
- don't require aws credential args, to allow using iam instance profiles
- fix bug where if an exception was raised in the watch method the sleep period got skipped. Better to wait the usual interval before retrying than to start running the failing discover instances method every few ms if something is wrong.

### DIFF
--- a/lib/synapse/service_watcher/ec2tag.rb
+++ b/lib/synapse/service_watcher/ec2tag.rb
@@ -45,11 +45,14 @@ class Synapse::ServiceWatcher
         raise ArgumentError, "Invalid server_port_override value"
       end
 
-      # Required, but can use well-known environment variables.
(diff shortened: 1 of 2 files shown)


jolynch:
@dcosson This looks good to merge to me. Can you rebase for merge (I think it's just an issue with the Gemfile.lock)?

From my understanding this will close #98 and #84.

dcosson:
@jolynch yup. Just rebased so I think it's good to go.

jolynch:
@dcosson Thanks for the PR! Please let me know if latest master works for you with the dynamic watcher creation and this patch (we don't use the ec2tag watcher so I am not sure it works in production heh).

dcosson:
:+1: Thanks for merging

PR #128, opened 2015-07-20
Sorting backends

### LINKED ISSUE #127: Why backends are shuffled?
Digging into the code I found out that the backend order is shuffled, instead of being consistent for every node. This is causing me troubles when balancing the requests via consistent hash.

See: https://github.com/airbnb/synapse/blob/master/lib/synapse/haproxy.rb#L665

asiragusa:
Fixes #127

### DIFF
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -662,7 +662,7 @@ module Synapse
       stanza = [
         "\nbackend #{watcher.name}",
         config.map {|c| "\t#{c}"},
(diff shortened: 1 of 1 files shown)


jolynch:
I think per the discussion in https://github.com/airbnb/synapse/pull/130#discussion_r35285222 indicates that this might break core functionality for airbnb. Perhaps this can be an option to either sort or not sort the backends?

jolynch:
@asiragusa If you still need this please re-open this PR with the functionality exposed as an option for backends.

PR #129, opened 2015-07-21
Test synapse on multiple versions of Ruby

rickypai:


igor47:
:+1: -- after this is merged, we should release a new version of synapse and roll it out in our infra

PR #130, opened 2015-07-22
Merge the forks

jolynch:
I believe that this takes your feedback from https://github.com/airbnb/synapse/pull/115, adds some documentation of the features we added.

Major changes in this fork:
- Service watchers that use the same zookeeper hosts will share one underlying zookeeper connection
- Allow output of registrations to json files on the filesystem instead of just HAProxy
- Allow rate limiting HAProxy restarts more aggressively than socket reconfigures
- Add a cache of recently seen backends that are put in "disabled" and can then be enabled over the socket without a restart.


README.md
```diff
@@ -198,6 +198,19 @@ If you do not list any default servers, no proxy will be created.  The
 `default_servers` will also be used in addition to discovered servers if the
 `keep_default_servers` option is set.
 
+If you do not list any `default_servers`, and all backends for a service
+disappear then the previous known backends will be used.  Disable this behavior
+by unsetting `use_previous_backends`.
```
igor47: oh thank god, finally!

lib/synapse/file_output.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def write_backends_to_file(service_name, new_backends)
+      data_path = File.join(@opts['output_directory'], "#{service_name}.json")
+      begin
+        old_backends = JSON.load(File.read(data_path))
+      rescue Errno::ENOENT
+        old_backends = nil
+      end
+
+      if old_backends == new_backends
```
igor47: a comment explaining this might be helpful to avoid refactoring this away in the future. are you just trying to avoid modified timestamps/inotify events on the files?

lib/synapse/haproxy.rb
```diff
@@ -655,18 +693,33 @@ def generate_frontend_stanza(watcher, config)
...
-        log.warn "synapse: no backends found for watcher #{watcher.name}"
+        log.debug "synapse: no backends found for watcher #{watcher.name}"
       end
 
       stanza = [
         "\nbackend #{watcher.name}",
         config.map {|c| "\t#{c}"},
-        watcher.backends.shuffle.map {|backend|
-          backend_name = construct_name(backend)
+        backends.map {|backend_name, backend|
```
igor47: curious why you got rid of the `shuffle` here; doesn't it helps to distribute the load more evenly? if all consumers have the same ordering for a set of backends and are using the `roundrobin` algorithm, does that mean that all consumers will first contact the first backend for the first request?

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -54,7 +59,7 @@ def discover
         node = @zk.get("#{@discovery['path']}/#{id}")
 
         begin
-          host, port, name = deserialize_service_instance(node.first)
+          host, port, name, weight = deserialize_service_instance(node.first)
```
igor47: i'm guessing that you're using `weight` to preferentially route within latency zones, but i don't see this attribute used anywhere else. is this branch still missing some code?

igor47:
looks pretty good overall. i left a few questions in a few places. i also think the test coverage is a little light, especially around zk connection pooling and all of the new haproxy functionality.

i think a few additional airbnb peeps from our production infrastructure team would like to take a look. ping @schleyfox @jtai 

thanks for doing this, i'm excited about all of the great new functionality (especially connection pooling, the disabling of the keep-backends misfeature, and the haproxy state file so you can continue to see the disabled backends)

lib/synapse/file_output.rb
jolynch: Yea, in case libraries watch these files for changes and react accordingly. I'll add a comment

lib/synapse/haproxy.rb
jolynch: This is true if all HAProxy instances start up at exactly the same moment and request loads are such that you notice this. In our experience things even out on their own (both haproxy reloads take time and request loads to servers has jitter). We much more often run into situations where HAProxy is restarting a lot and the HAProxy stats page becomes really hard to read which can impede debugging efforts.

This is purely so that the stats page across reloads doesn't change a ton. I can change it back if you prefer the random behavior :-)

lib/synapse/service_watcher/zookeeper.rb
jolynch: Yea, we're working on exposing this to HAProxy. This is the first step (see https://github.com/Yelp/synapse/pull/14 for discussion between the author and me regarding this)

jolynch:
Yea, the connection pooling and haproxy cache code was sorta done under P0 pressure (site was actively degraded kind of P0), so we didn't write too many tests heh. Do you think the lack of test coverage blocks merge or is it something that I can work on in separate PRs against this repo after the fork is gone?

lib/synapse/haproxy.rb
jolynch: Also possibly helps with issue #127

schleyfox: I think the shuffling is helpful for some of the things that we have that have a predictable number (per box) of long lived connections like our MySQL databases.  Especially since these connections survive haproxy restarts and the time of reconnection is very synchronized on every server, I could definitely see this causing problems for our use case.

This may help with #127, but unless I'm missing something, I don't understand how consistent hashing could be expected to play nicely with synapse/haproxy.

jolynch: That seems like a compelling use case. Going forward maybe we can add an option if people want it to be sorted vs randomized, but I think it makes sense to do that in future versions.

schleyfox:
I gave it a skim and it seems pretty reasonable to me. I'm excited.

lib/synapse/haproxy.rb
```diff
@@ -43,6 +45,7 @@ class Haproxy
         "option abortonclose",
         "option accept-invalid-http-response",
         "option allbackups",
+        "option allredisp",
```
jtai: Will this cause problems for standard builds of haproxy?

lib/synapse/service_watcher/base.rb
```diff
@@ -95,13 +100,45 @@ def validate_discovery_opts
...
+
+      if new_backends.to_set == @backends.to_set
+        return false
+      end
+
+      if new_backends.empty?
+        if @default_servers.empty?
+          if @use_previous_backends
+            # Discard this update
+            log.warn "synapse: no default servers for service #{@name};" \
```
jtai: Should this (and the next `warn` statement) also mention that there are no backends for the service? E.g., "synapse: no backends for service #{@name} and no default servers configured; using previous backends: #{backends.inspect}"

lib/synapse/haproxy.rb
igor47: no this is just a list of valid options so we can validate the in-synapse haproxy config. we already have a few options here which won't work with haproxy < 1.5; the symptoms if you use unsupported options is that synapse will start, but it's every attempt to restart haproxy will fail

jtai:
These improvements are awesome! Thank you!

lib/synapse/haproxy.rb
jolynch: Both my patches were accepted into 1.6, and Willy just released 1.6-dev3 so we're almost off our fork! Once we upgrade to 1.6 I can go through and re-sync this list with what haproxy accepts. There might be some backwards incompatible changes in 1.6 from 1.5 (only one I have ran into was listen no longer accepts ports in the listen line) so that should be a fun exercise.

lib/synapse/service_watcher/base.rb
jolynch: Sure, seems clearer.

igor47:
okay i think this is ready to merge. once that's done, i'll bump the version and release a build, and we can start trying to run it in airbnbland

jolynch:
We're going to start dogfooding 3e36e88559 and then when you release we'll switch to that.

igor47:
ok i released v0.12.1; i'm going to start trying it here as well

lib/synapse/haproxy.rb
asiragusa: @schleyfox It doesn't really work nicely, it works as just hashing. 

In order to work better with consistent hashing we may keep dead backends for longer.

This may be done by creating two keys in zk, one listing all the machines in the pool and the other telling if it's alive. The key listing all the machines may have an adjustable timeout and the default value should be low to work nicely everywhere.

However in case the ordering is really important, we may also introduce sequential numbering of nodes and if an hole is detected in the middle we may fill it with a fake machine. This fake machine will be replaced once a new one joins the cluster.

Btw I'm ok with making shuffling the default option, as long as this is a very specific use case.

OT: take a look at https://github.com/dockerparis/DockStack, it makes really easy to use nerve/synapse as a Docker ambassador.

igor47:
my smoke testing went well, and now i'm allowing synapse v0.12.1 to propagate through our infrastructure; i'm going to tackle airbnb/nerve#71 now

PR #131, opened 2015-07-28
Weights in generated config

bobtfish:
cc @vulpine

Rebased against airbnb/synapse master

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ This section is its own hash, which should contain the following keys:
 * `backend`: additional lines passed to the HAProxy config in the `backend` stanza of this service
 * `listen`: these lines will be parsed and placed in the correct `frontend`/`backend` section as applicable; you can put lines which are the same for the frontend and backend here.
 * `shared_frontend`: optional: haproxy configuration directives for a shared http frontend (see below)
+* `ignore_weights`: optional: stops haproxy backend 'weight' options being generated, even if the Nerve registrations contain this information. This will cause all backend servers to be treated equally by haproxy.
 
 ### Configuring HAProxy ###
 
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -533,6 +533,7 @@ module Synapse
       @opts['do_writes'] = true unless @opts.key?('do_writes')
       @opts['do_socket'] = true unless @opts.key?('do_socket')
       @opts['do_reloads'] = true unless @opts.key?('do_reloads')
+      @opts['ignore_weights'] = true if @opts.key?('ignore_weights') and @opts['ignore_weights']
 
       # how to restart haproxy
       @restart_interval = @opts.fetch('restart_interval', 2).to_i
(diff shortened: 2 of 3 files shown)


lib/synapse/haproxy.rb
```diff
@@ -719,6 +720,7 @@ def generate_backend_stanza(watcher, config)
           backend = backends[backend_name]
           b = "\tserver #{backend_name} #{backend['host']}:#{backend['port']}"
           b = "#{b} cookie #{backend_name}" unless config.include?('mode tcp')
+          b = "#{b} weight #{backend['weight']}" if backend['weight'] && backend['weight'].is_a?(Fixnum) and backend['weight'] > 0 and !@opts['ignore_weights']
```
jolynch: I don't think we want to ignore 0 weight backends because that seems like a reasonable use case? (I believe that nerve supplies nil if there is no weight).

What do you think?

README.md
```diff
@@ -222,6 +222,7 @@ This section is its own hash, which should contain the following keys:
 * `backend`: additional lines passed to the HAProxy config in the `backend` stanza of this service
 * `listen`: these lines will be parsed and placed in the correct `frontend`/`backend` section as applicable; you can put lines which are the same for the frontend and backend here.
 * `shared_frontend`: optional: haproxy configuration directives for a shared http frontend (see below)
+* `ignore_weights`: optional: stops haproxy backend 'weight' options being generated, even if the Nerve registrations contain this information. This will cause all backend servers to be treated equally by haproxy.
```
jolynch: Can you mention what the default is?

jolynch:
Looks reasonable, can you check that this works with the state file (we should/will add tests of that soon)?

Also this doesn't have code for doing updates over the socket, but we presumably want that. Is that going to come in a follow up patch?

spec/lib/synapse/haproxy_spec.rb
```diff
@@ -29,4 +32,14 @@ class MockWatcher; end;
     expect(subject.generate_backend_stanza(mockwatcher, mockConfig)).to eql(["\nbackend example_service", ["\tmode tcp"], ["\tserver somehost:5555 somehost:5555 check inter 2000 rise 3 fall 2"]])
   end
 
+  it 'generates backend stanza with weight' do
+    mockConfig = []
+    expect(subject.generate_backend_stanza(createmockwatcher([{ 'weight' => 1, 'host' => 'somehost', 'port' => '5555'}]), mockConfig)).to eql(["\nbackend example_service", [], ["\tserver somehost:5555 somehost:5555 cookie somehost:5555 weight 1 check inter 2000 rise 3 fall 2"]])
+  end
+
+  it 'generates backend stanza withour bad weight' do
```
vulpine: Minor typo here.

bobtfish:
@jolynch ack - thanks for the feedback. I'll add tests around the state file, and I had totally forgotten the weight updates over the socket. I'd however prefer to do that as a separate patch to keep things relatively simple. Does that work for you?

jolynch:
My only issue with doing the socket update in a subsequent change is that something else will have to trigger a HAProxy reload to pick up any change in weights (since this patch neither informs the haproxy module that it needs to reload nor changes the weights over the stats socket). At the very least I think it should inform haproxy to restart?

jolynch:
@bobtfish #100 seems rather similar to this, can we sync up about how we want to be able to add metadata to nodes? I'm not against adding a bunch of top level keys, but I think we should at least have a plan.

jolynch:
#140 shows a technique for noticing things have changed via the state file. Perhaps we should just make the state file mandatory ...

PR #132, opened 2015-08-12
Allow pluggable watchers

jolynch:
Similar to what @bobtfish did in his nerve commit 258d62a823, this commit
allows Smartstack consumers to provide their own pluggable service
watchers.

Simply name your class correctly and put it in the right place and poof,
you have your own service watcher.

### DIFF
--- a/lib/synapse/service_watcher/README.md
+++ b/lib/synapse/service_watcher/README.md
@@ -1,0 +1,58 @@
+Please, do not submit pull requests with new watchers, as Synapse and Nerve
+both support a plugin system for adding watchers and reporters respectively.
+
+## Watcher Classes
+
+Watchers are the piece of Synapse that watch an external service registry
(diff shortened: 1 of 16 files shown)


jolynch:
@igor47 We should probably create a repo for community contributed synapse watchers and nerve reporters. Also I need to actually figure out how to ninja in a watcher/reporter class without a checkout + patch.

lib/synapse/service_watcher/README.md
```diff
@@ -0,0 +1,58 @@
+Please, do not submit pull requests with new watchers, as Synapse and Nerve
```
jolynch: Let's give folks somewhere to contribute these though. Don't want to discourage releasing watchers.

bobtfish:
:+1:

jolynch:
Rebased and changed the documentation to not mention the yet to be contrib repo.

PR #133, opened 2015-08-12
Support Netflix Exhibitor

lsjostro:
This PR will add support for Netflix Exhibitor. It's more or less a copy of the regular zookeeper watcher adding but adds polling of the Exhibitor REST endpoint for zk ensemble changes. If the ensemble changes this watcher will detect that and stop the watcher.

This is great when maintaining zookeeper in the cloud to not have static zookeeper endpoints.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,4 +5,3 @@ RSpec::Core::RakeTask.new(:spec)
 
 task :test => :spec
 task :default => :spec
-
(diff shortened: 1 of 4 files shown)


igor47:
i left some feedback on the corresponding nerve PR (airbnb/nerve#74) that i would like to see answered before reviewing this

PR #135, opened 2015-08-15
Explicit watcher haproxy backend names

jolynch:
This allows acls in frontend sections to reference backends without coupling
to the current convention of naming backends after watcher names

We're interested in doing farm failover at Yelp as described in http://blog.haproxy.com/2013/12/23/failover-and-worst-case-management-with-haproxy/ and I think all we need to do this is to have backup service watchers that have no port (so they'll get no frontend) and then have acls in the main service watchers that redirect to the backup backends if needed. However, to do this we need to know the names of backends and I'd rather not couple our configuration to something that synapse reserves the right to change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,8 @@ This section is its own hash, which should contain the following keys:
 * `server_options`: the haproxy options for each `server` line of the service in HAProxy config; it may be left out.
 * `frontend`: additional lines passed to the HAProxy config in the `frontend` stanza of this service
 * `backend`: additional lines passed to the HAProxy config in the `backend` stanza of this service
+* `backend_name`: The name of the generated HAProxy backend for this service
+  (defaults to the service's key in the `services` section)
 * `listen`: these lines will be parsed and placed in the correct `frontend`/`backend` section as applicable; you can put lines which are the same for the frontend and backend here.
 * `shared_frontend`: optional: haproxy configuration directives for a shared http frontend (see below)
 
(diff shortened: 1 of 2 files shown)


jolynch:
For context I'm thinking something like:

```json
  "services": {                                                                 
    "service1": {                                                               
      "default_servers": [],                                                    
      "use_previous_backends": false,                                           
      "discovery": {                                                            
        "method": "zookeeper",                                                  
        "path": "/services/service1",                                           
        "hosts": [                                                              
          "localhost:2181"                                                      
        ]                                                                       
      },                                                                        
      "haproxy": {                                                              
        "port": 3213,                                                           
        "backend_name": "service1",                                             
        "listen": [                                                             
          "mode http",                                                          
        ],                                                                      
        "frontend": [                                                           
          "acl needs_fallback nbsrv(service1) le 0",                            
          "use_backend service1_backup if needs_fallback"                       
        ]                                                                       
      }                                                                         
    },                                                                          
    "service1_backup": {                                                        
      "default_servers": [],                                                    
      "use_previous_backends": false,                                           
      "discovery": {                                                            
        "method": "zookeeper",                                                  
        "path": "/services/service1_backup",                                    
        "hosts": [                                                              
          "localhost:2181"                                                      
        ]                                                                       
      },                                                                        
      "haproxy": {                                                              
        "backend_name": "service1_backup",                                                                                                   
      }                                                                         
    }
}
```

jolynch:
fwiw the shared frontend section sorta allows this although afaik it only allows one shared frontend for the entire haproxy instance?

jnb:
:+1:

PR #136, opened 2015-08-15
Some minor documentation fixups

jolynch:
Reading through the readme is a bit confusing as to which haproxy is which and where file_output actually goes, I've tried fixing that.


igor47:
:+1:

PR #140, opened 2015-10-03
Allow arbitrary haproxy_server_options to be supplied to HAProxy.

jolynch:
This allows operators to declare that particular backends are backups, should have different failure modes (fall/rise), healthcheck rates, etc ... We need to add support to nerve to be able to pass these through, but this should help with #100.

If you want restarts to be forced when the state changes, you must use
the state file.

### DIFF
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -704,6 +704,16 @@ module Synapse
       # setting the enabled state.
       watcher.backends.each do |backend|
         backend_name = construct_name(backend)
+        # If we have information in the state file that allows us to detect
+        # server option changes, use that to potentially force a restart
+        if backends.has_key?(backend_name)
+          old_backend = backends[backend_name]
+          if (old_backend.fetch('haproxy_server_options', "") !=
+              backend.fetch('haproxy_server_options', ""))
+            log.info "synapse: restart required because haproxy_server_options changed for #{backend_name}"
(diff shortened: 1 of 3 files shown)


Jaykah:
That was fast, thanks! Was just about to address this. 

Have tested this and can confirm that it is working as expected with all of the options I need.

Also, worth noting that if someone wants to do http checks on custom ports, they will have to add       - ```"option httpchk /"``` to haproxy section, in addition to ```haproxy_server_options: "check port 9203"```

jnb:
:+1: 

I like using the state file instead of querying haproxy.  In the future I think it might be worth abstracting this component out some more.

Jaykah:
Just tested https://github.com/airbnb/nerve/pull/75 and wanted to follow up on this pull request - for some reason, when Synapse is getting data from Zookeeper that includes haproxy_server_options it does not use them, generating a Haproxy config without those options.

jolynch:
Did you enable the state file?

Jaykah:
@jolynch No, state cache is not enabled.

jolynch:
@Jaykah Heh I forgot to merge #144 . I've merged it now and can start testing on our infra next week (after thanksgiving).

Remember to enable the `state_file_path` from https://github.com/airbnb/synapse#configuring-haproxy so that when the haproxy server options change synapse will restart haproxy.

PR #141, opened 2015-10-05
Update README.md with better installation directions

tdooner:
The previous installation directions are not great -- they are ambiguous
between whether synapse should be used as a library (and placed in the
depending app's Gemfile) vs being used as a standalone application (and
installed via `gem install`).

The best way I know to deploy a Ruby application is to use `gem install
--install-dir [path]` to install the application in that path. This
allows:

* the gem to be installed by any user, not just root
* the version of synapse's dependencies to be isolated from other system
  gems
* easier redistribution as a system package

The downside is a relative large disk footprint - 36 MB as of writing.
Many files could be excluded from a system package, like the cache/
directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,19 +92,26 @@ HAProxy will be transparently reloaded, and your application will keep running w
 
 ## Installation
 
-Add this line to your application's Gemfile:
+To download and run the synapse binary, first install a version of ruby.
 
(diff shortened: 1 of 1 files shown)


jolynch:
Yea, this looks like a better guide, I'll add a note about 1.9 compat.

PR #143, opened 2015-10-13
Cleanup old files that synapse no longer manages

### LINKED ISSUE #139: File output does not purge unknown services
I came across a fun bug today where we removed a service and synapse decided to stop managing the output file for us. This is somewhat surprising and we should probably purge any service files we don't recognize.

jolynch:
It turns out that when removing services these files stick around
because synapse stops managing them, so in this patch we clean up
anything in the output directory that shouldn't be there.

Should fix #139

Also adds tests of the file output generator.

### DIFF
--- a/lib/synapse/file_output.rb
+++ b/lib/synapse/file_output.rb
@@ -28,6 +28,7 @@ module Synapse
       watchers.each do |watcher|
         write_backends_to_file(watcher.name, watcher.backends)
       end
+      clean_old_watchers(watchers)
     end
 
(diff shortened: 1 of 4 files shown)


jolynch:
Potentially a lot of file system IO going on here, but I think we're already reading all the files off disk and parsing them ... so ... this is probably not so bad.

Maybe we should throw these things in threads or some such...

igor47:
lgtm! thanks for adding specs!

PR #144, opened 2015-10-14
Adds support to the zookeeper service watcher for haproxy_server_options

jolynch:
This is such a bad design, I really dislike that we munge the zk data
multiple times. Unfortunately I don't know a better way to do it yet. I
will work on it.


jolynch:
I did some manual tests, but it would be great if we had zk specs at some point ...

PR #146, opened 2015-10-29
Use environment variable SYNAPSE_ZK_ROOT as an optional prefix to Zoo…

ianbamforth:
…keeper service paths

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -27,7 +27,6 @@ GEM
     excon (0.45.4)
     ffi (1.9.3-java)
     json (1.8.3)
-    json (1.8.3-java)
     little-plugger (1.1.3)
     logging (1.8.2)
       little-plugger (>= 1.1.3)
@@ -37,7 +36,6 @@ GEM
     multi_json (1.11.2)
     nokogiri (1.6.6.2)
       mini_portile (~> 0.6.0)
-    nokogiri (1.6.6.2-java)
     pry (0.9.12.2)
       coderay (~> 1.0.5)
       method_source (~> 0.8)
@@ -64,8 +62,6 @@ GEM
     rspec-support (3.1.2)
     safe_yaml (1.0.4)
     slop (3.4.6)
-    slyphon-log4j (1.2.15)
-    slyphon-zookeeper_jar (3.3.5-java)
     spoon (0.0.4)
       ffi
     webmock (1.18.0)
@@ -92,4 +88,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   1.10.6
+   1.10.5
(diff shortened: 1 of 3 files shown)


jolynch:
What is the advantage of this over configuring the path differently in your synapse.conf.json? 

I'm not sure I see the benefit?

ianbamforth:
I’m using AWS, spinning up new instances from AMIs, and I want to be able to use the exact same image in staging / test / production.  The only difference between the servers will be user-data passed in at boot time, which I hope to use to allow services to discover what environment they have been launched in.

 

I could take the approach of creating a different image for each of staging / test / production (and any number of others I guess!) but it feels wrong and if they differ at all then you lose a degree of confidence in any testing you perform.

 

I’ve prototyped this idea, and it seems to be working ok.  It means I can spin up a server with user-data env=staging, perform a bunch of tests, reboot with user-data env=production (or more likely just launch another copy).  Seems like a nice way to use auto discovery in synapse to separate the two environments.

 

Ian

jolynch:
@ianbamforth But what happens when you need to add a new service or change the HAProxy parameters like timeouts? You'll re-deploy your service AMIs? I generally feel like trying to bake Synapse into an AMI without some cron job or something that automatically reconfigures Synapse is going to lead to some difficulty down the road.

If you're just trying to get a proof of concept going maybe a script that reads environment variables and writes them into Synapse configuration files before starting Synapse?

I guess what I'm trying to get at is that the Synapse configuration file is the way to configure Synapse precisely because files are so flexible. We can have cron jobs that regenerate them, or one can do something fancy like setup a process that watches some source of truth (e.g. a git repo or zookeeper itself) and regenerates the config any time that changes, or we can manifest environment variables into it.

Jaykah:
Another option you can leverage (and the one that we are currently using in production) is passing those deployment type tags as parameters into Docker containers that have start scripts as entrypoints. Then, in the aforementioned scripts you can directly sed pre-defined placeholders with the deployment types that you pass (thus adjusting the Synapse config directly). That way when Synapse launches it starts working in the correct environment right away without the need to restart/reload.

ianbamforth:
I had considered using a user-data script (we're not currently using Docker) to do that kind of thing.  The issue is that we have multiple different services written in a variety of languages, self-advertising (mostly without Nerve), configured in a variety of ways, so the script would have to handle not just updating the Synapse config but also locating all of them, which would be more complex.

The change I made works quite well, but I've been having trouble getting an Upstart script to read the user-data from 169.254.169.254 - issues with forking I think.

I may consider having a user data script that puts the root in e.g. /etc/environment, which would then be easy to access from any interested process.

jolynch:
@ianbamforth, I really prefer if the mechanism for configuring synapse remains the configuration file. If you need help getting a user data script that fills in a template or some such let me know, happy to help.

jolynch:
I'm going to close this because I think it's adding complexity to solve a problem that can be solved well with the existing configuration systems.

PR #148, opened 2015-11-25
add apache aurora service discovery support - issues/147

JustinVenus:
Adding support for aurora style service discovery.

Without patch:
```shell
[root@c90f0a9b92d9 /]# /usr/local/bin/synapse -c /etc/synapse/synapse.json
W, [2015-11-25T16:54:33.707037 #56]  WARN -- Synapse::ZookeeperWatcher: synapse: service hello_world: haproxy config does not include a port; only backend sections for the service will be created; you must move traffic there manually using configuration in `extra_sections`
I, [2015-11-25T16:54:33.707253 #56]  INFO -- Synapse::Synapse: synapse: starting...
I, [2015-11-25T16:54:33.707291 #56]  INFO -- Synapse::ZookeeperWatcher: synapse: starting ZK watcher hello_world @ hosts: 10.150.150.224:2181, path: /aurora/nobody/devel/hello_world
I, [2015-11-25T16:54:33.707315 #56]  INFO -- Synapse::ZookeeperWatcher: synapse: zookeeper watcher connecting to ZK at 10.150.150.224:2181
I, [2015-11-25T16:54:33.707337 #56]  INFO -- Synapse::ZookeeperWatcher: synapse: creating pooled connection to 10.150.150.224:2181
I, [2015-11-25T16:54:33.829500 #56]  INFO -- Synapse::ZookeeperWatcher: synapse: successfully created zk connection to 10.150.150.224:2181
I, [2015-11-25T16:54:33.829642 #56]  INFO -- Synapse::ZookeeperWatcher: synapse: retrieved zk connection to 10.150.150.224:2181
I, [2015-11-25T16:54:34.051414 #56]  INFO -- Synapse::ZookeeperWatcher: synapse: discovering backends for service hello_world
E, [2015-11-25T16:54:34.219371 #56] ERROR -- Synapse::ZookeeperWatcher: synapse: invalid data in ZK node member_0000000028 at /aurora/nobody/devel/hello_world: uninitialized constant Synapse::ZookeeperWatcher::ValueError
E, [2015-11-25T16:54:34.291087 #56] ERROR -- Synapse::ZookeeperWatcher: synapse: invalid data in ZK node member_0000000036 at /aurora/nobody/devel/hello_world: uninitialized constant Synapse::ZookeeperWatcher::ValueError
E, [2015-11-25T16:54:34.363133 #56] ERROR -- Synapse::ZookeeperWatcher: synapse: invalid data in ZK node member_0000000027 at /aurora/nobody/devel/hello_world: uninitialized constant Synapse::ZookeeperWatcher::ValueError
E, [2015-11-25T16:54:34.434785 #56] ERROR -- Synapse::ZookeeperWatcher: synapse: invalid data in ZK node member_0000000031 at /aurora/nobody/devel/hello_world: uninitialized constant Synapse::ZookeeperWatcher::ValueError
E, [2015-11-25T16:54:34.526296 #56] ERROR -- Synapse::ZookeeperWatcher: synapse: invalid data in ZK node member_0000000030 at /aurora/nobody/devel/hello_world: uninitialized constant Synapse::ZookeeperWatcher::ValueError
E, [2015-11-25T16:54:34.598868 #56] ERROR -- Synapse::ZookeeperWatcher: synapse: invalid data in ZK node member_0000000035 at /aurora/nobody/devel/hello_world: uninitialized constant Synapse::ZookeeperWatcher::ValueError
E, [2015-11-25T16:54:34.675987 #56] ERROR -- Synapse::ZookeeperWatcher: synapse: invalid data in ZK node member_0000000034 at /aurora/nobody/devel/hello_world: uninitialized constant Synapse::ZookeeperWatcher::ValueError
E, [2015-11-25T16:54:34.752758 #56] ERROR -- Synapse::ZookeeperWatcher: synapse: invalid data in ZK node member_0000000033 at /aurora/nobody/devel/hello_world: uninitialized constant Synapse::ZookeeperWatcher::ValueError
E, [2015-11-25T16:54:34.827884 #56] ERROR -- Synapse::ZookeeperWatcher: synapse: invalid data in ZK node member_0000000032 at /aurora/nobody/devel/hello_world: uninitialized constant Synapse::ZookeeperWatcher::ValueError
```

With patch:
```shell
[root@c90f0a9b92d9 /]# /usr/local/bin/synapse -c /etc/synapse/synapse.json
W, [2015-11-25T16:42:28.500523 #44]  WARN -- Synapse::ZookeeperWatcher: synapse: service hello_world: haproxy config does not include a port; only backend sections for the service will be created; you must move traffic there manually using configuration in `extra_sections`
I, [2015-11-25T16:42:28.500835 #44]  INFO -- Synapse::Synapse: synapse: starting...
I, [2015-11-25T16:42:28.500954 #44]  INFO -- Synapse::ZookeeperWatcher: synapse: starting ZK watcher hello_world @ hosts: 10.150.150.224:2181, path: /aurora/nobody/devel/hello_world
I, [2015-11-25T16:42:28.501033 #44]  INFO -- Synapse::ZookeeperWatcher: synapse: zookeeper watcher connecting to ZK at 10.150.150.224:2181
I, [2015-11-25T16:42:28.501109 #44]  INFO -- Synapse::ZookeeperWatcher: synapse: creating pooled connection to 10.150.150.224:2181
I, [2015-11-25T16:42:28.587700 #44]  INFO -- Synapse::ZookeeperWatcher: synapse: successfully created zk connection to 10.150.150.224:2181
I, [2015-11-25T16:42:28.587919 #44]  INFO -- Synapse::ZookeeperWatcher: synapse: retrieved zk connection to 10.150.150.224:2181
I, [2015-11-25T16:42:28.827093 #44]  INFO -- Synapse::ZookeeperWatcher: synapse: discovering backends for service hello_world
I, [2015-11-25T16:42:29.833430 #44]  INFO -- Synapse::ZookeeperWatcher: synapse: discovered 9 backends for service hello_world
I, [2015-11-25T16:42:29.833840 #44]  INFO -- Synapse::Synapse: synapse: configuring haproxy
```

### DIFF
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -41,6 +41,7 @@ class Synapse::ServiceWatcher
         unless @discovery['hosts']
       raise ArgumentError, "invalid zookeeper path for service #{@name}" \
         unless @discovery['path']
+      @decode_method = JSON.parse
       if @discovery['decode']
         raise ArgumentError, "missing or invalid decode method #{@discovery['decode']}" \
           unless @discovery['decode']['method']
@@ -52,8 +53,6 @@ class Synapse::ServiceWatcher
               JSON.parse(data)['serviceEndpoint']
             end
           end
-        else
-          @decode_method = JSON.parse
         end
       end
     end
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -41,6 +41,21 @@ class Synapse::ServiceWatcher
         unless @discovery['hosts']
       raise ArgumentError, "invalid zookeeper path for service #{@name}" \
         unless @discovery['path']
+      if @discovery['decode']
+        raise ArgumentError, "missing or invalid decode method #{@discovery['decode']}" \
+          unless @discovery['decode']['method']
+        if @discovery['decode']['method'] == 'aurora'
+          @decode_method = lambda do |data|
+            if @discovery['decode']['port_name']
+              JSON.parse(data)['additionalEndpoints'][@discovery['decode']['port_name']]
+            else
+              JSON.parse(data)['serviceEndpoint']
+            end
+          end
+        else
+          @decode_method = JSON.parse
+        end
+      end
     end
 
     # helper method that ensures that the discovery path exists
@@ -173,12 +188,8 @@ class Synapse::ServiceWatcher
     # decode the data at a zookeeper endpoint
     def deserialize_service_instance(data)
       log.debug "synapse: deserializing process data"
-      decoded = JSON.parse(data)
+      decoded = @decode_method.call(data)
 
-      # supporting Apache Aurora style service discovery
-      if decoded.include? 'serviceEndpoint' and decoded.include? 'status' and decoded['status'] == 'ALIVE'
-        decoded = decoded['serviceEndpoint']
-      end
       host = decoded['host'] || (raise ValueError, 'instance json data does not have host key')
       port = decoded['port'] || (raise ValueError, 'instance json data does not have port key')
       name = decoded['name'] || nil
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -175,6 +175,10 @@ class Synapse::ServiceWatcher
       log.debug "synapse: deserializing process data"
       decoded = JSON.parse(data)
 
+      # supporting Apache Aurora style service discovery
+      if decoded.include? 'serviceEndpoint' and decoded.include? 'status' and decoded['status'] == 'ALIVE'
+        decoded = decoded['serviceEndpoint']
+      end
       host = decoded['host'] || (raise ValueError, 'instance json data does not have host key')
       port = decoded['port'] || (raise ValueError, 'instance json data does not have port key')
       name = decoded['name'] || nil


jolynch:
@JustinVenus 
Given my understanding a of aurora registrations from #104 a single service can have multiple ports, which we might want to register separately?

What are your thoughts on having a separate watcher that inherits from the zookeeper watcher? I'm slightly partial to keeping it separate and overriding the validation and deserialization.

JustinVenus:
I'm probably biased on how I think aurora's service discovery should be used.  IMO a "serviceEndpoint" should be the only thing exposed to other services, b/c it is advertising the main port for the service.

Yes aurora has "additionalEndpoints".  I personally use "additionalEndpoint" ports for things like the "health" or "aurora" port which aurora expects to use, but you'd never want to expose on a well known port.  Other ports might be something like JMX access, which falls into you probably don't want to make that easily accessible.

Initially I went down the path of making another watcher to support the "additionalEndpoints", but I could not find a reasonable use case to support this in an aurora/mesos cluster.  The amount of code to support the change for aurora (that fits into my view on usage) ended up being trivial and IMO not worth separation into yet another watcher.

Also as a side note, I think zookeeper_dns get's my enhancement for free.

jolynch:
I can see why the other PR want support for additional ports, but I also understand that a whole new class might be overkill.

What if instead the zookeeper watcher took a decode argument which defaults to a dictionary containing {'method': 'nerve'}, and can optionally be 'aurora'? Then we could just have port options on the Aurora parsing.

JustinVenus:
I think I see where you are going w/ this and it seems extensible for the next service discovery thing that uses zookeeper.  How would the decode option be passed to the watcher from config?

jolynch:
@JustinVenus It would just be a top level argument that the zookeeper watcher accepts and validates in the service config (e.g. same as https://github.com/airbnb/synapse#zookeeper). It would have to be a hash so that we can pass config deeper. For example if you add an option called "decode" which would know what to do with ``{'method': 'nerve'}`` or ``{'method': 'aurora', 'port_name': 'foobar'}``

JustinVenus:
@jolynch would something like this be reasonable?
```
[root@32622e27341e /]# diff -u /usr/local/share/gems/gems/synapse-0.12.1/lib/synapse/service_watcher/zookeeper.rb.orig /usr/local/share/gems/gems/synapse-0.12.1/lib/synapse/service_watcher/zookeeper.rb
--- /usr/local/share/gems/gems/synapse-0.12.1/lib/synapse/service_watcher/zookeeper.rb.orig	2015-11-25 17:13:20.000000000 +0000
+++ /usr/local/share/gems/gems/synapse-0.12.1/lib/synapse/service_watcher/zookeeper.rb	2015-11-30 21:38:47.497608224 +0000
@@ -41,6 +41,21 @@
         unless @discovery['hosts']
       raise ArgumentError, "invalid zookeeper path for service #{@name}" \
         unless @discovery['path']
+      if @discovery['decode']
+        raise ArgumentError, "missing or invalid decode method #{@discovery['decode']}" \
+          unless @discovery['decode']['method']
+        if @discovery['decode']['method'] == 'aurora'
+          @decode_method = lambda do |data|
+            if @discovery['decode']['port_name']
+              JSON.parse(data)['additionalEndpoints'][@discovery['decode']['port_name']]
+            else
+              JSON.parse(data)['serviceEndpoint']
+            end
+          end
+        else
+          @decode_method = JSON.parse
+        end
+      end
     end

     # helper method that ensures that the discovery path exists
@@ -169,7 +184,7 @@
     # decode the data at a zookeeper endpoint
     def deserialize_service_instance(data)
       log.debug "synapse: deserializing process data"
-      decoded = JSON.parse(data)
+      decoded = @decode_method.call(data)

       host = decoded['host'] || (raise ValueError, 'instance json data does not have host key')
       port = decoded['port'] || (raise ValueError, 'instance json data does not have port key')
```

jolynch:
@JustinVenus Yup that's basically what I'm thinking.

JustinVenus:
@jolynch I've updated the request based on guidance provided.

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -41,6 +41,21 @@ def validate_discovery_opts
         unless @discovery['hosts']
       raise ArgumentError, "invalid zookeeper path for service #{@name}" \
         unless @discovery['path']
+      if @discovery['decode']
```
jolynch: I think we need to set @decode_method even if we don't have a decode (i.e. the default is zookeeper)

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -41,6 +41,21 @@ def validate_discovery_opts
...
+        raise ArgumentError, "missing or invalid decode method #{@discovery['decode']}" \
+          unless @discovery['decode']['method']
+        if @discovery['decode']['method'] == 'aurora'
```
jolynch: I feel like this code should live in the constructor and be factored out into a function e.g. parse_decode_options or something. Feel free to go "meh" and I'll do it in a follow up.

JustinVenus:
@jolynch If it's all the same to you ("meh") I'd like to leave the refactoring to a follow up request.

jolynch:
Sure but we need to maintain backwards compatibility (set decode even us l if there is no config)

JustinVenus:
@jolynch backwards compatibility should be there.  AFAIK it is not an exception in ruby to access a non-existent key from a hash map.  Therefore the truth test should work and the default behavior is to use 'nerve' style service discovery.

```ruby
jvenus-signal-mbp13:~ jvenus$ irb
irb(main):001:0> foo = {'bar' => 'baz'}
=> {"bar"=>"baz"}
irb(main):002:0> foo['decode']
=> nil
```

jolynch:
@JustinVenus Yes if the decode param is provided then it is backwards compatible but it requires there to be a decode parameter where previously we did not. The else only applies to the ``if @discovery['decode']['method'] == 'aurora'`` I believe.

JustinVenus:
I'll put a guard in for ```@discovery['decode']['method']``` and update the request later today.

PR #149, opened 2015-11-30
Update docker-api

cap10morgan:
docker-api 1.7.x was generating 400 errors trying to talk to a Docker 1.7.x daemon. The latest version of docker-api (1.23.0) speaks a much newer version of the Docker Remote API and fixed this issue. The specs still pass w/ this newer version.

It appears that this used to rely on an implicit logging dependency pulled in by old versions of docker-api, but now it specifies the dependency explicitly because the latest docker-api seems to no longer pull it in.


jolynch:
@cap10morgan re-did this against master in #159. Let me know if that didn't work right :-)

PR #150, opened 2015-12-05
Reload if HAProxy config has been changed

che:


jolynch:
I'm a little confused by the purpose of this patch. Synapse already reloads haproxy if it can't make the stats socket update. What is your concern with that approach?

jolynch:
@che can you give more context?

che:
@jolynch you are right, but this is a fix for the situation when we need have only active servers in HAProxy config. If server is down - don't show this server in any case (for DOWN statuses).

Thanx!

jolynch:
@che the stats socket option should allow you to automatically down servers (put them in maint). Try setting ``socket_file_path `` and ``do_reloads ``, or am I missing something?

che:
@jolynch I changed solution(new commit) and renamed argument for more understanding.
Also current solution linked with https://github.com/airbnb/synapse/issues/94

Thanx!

jolynch:
@che I think that #94 has to do with a bad interaction in Docker, I don't believe that this change would fix it.

The current master branch of Synapse can automatically MAINT backends (bring them down). If you want to reload every time a server goes down instead of MAINTing then you turn set ``do_reloads`` to true and ``do_socket`` to false. If you want to use the stats socket to MAINT without reloading (preferable in the general case) then you set ``do_reloads`` and ``do_socket`` to true.

jolynch:
I guess to get to the root of this, what is the problem you are observing that you are trying to fix?

che:
>>> I guess to get to the root of this, what is the problem you are observing that you are trying to fix?

You are right, NET_ADMIN and qdisk solved our problems on Docker side.

che:
>>> The current master branch of Synapse can automatically MAINT backends (bring them down). If you want to reload every time a server goes down instead of MAINTing then you set do_reloads to true and do_socket to false. If you want to use the stats socket to MAINT without reloading (preferable in the general case) then you set do_reloads and do_socket to true.

We have in config:  do_writes:  true,  do_reloads: true,  do_socket:  true
We use HAProxy 1.5.15 and Synapse 0.12.2, but it doesn't work because if server is DOWN - the server still show as DOWN.
Our solution makes it possible to check the status of the server and if the server is not active (DOWM), then restart HAProxy then we will show only active (UP) servers.

jolynch:
@che Are there any warnings or errors on the command line? My understanding of the logic is that if it can't talk over the stats socket then it will initiate a reload, so I'm slightly puzzled by this situation.

Are you running this inside a docker container? If so what are you using as an init substitute?

jolynch:
@che bump

jolynch:
I'm going to close this because I don't think it's needed. Please re-open with more context if I'm wrong about that :-)

PR #152, opened 2015-12-22
Fix zookeeper discovery port override

hightoxicity:
Let server_port_override works for zookeeper discovery watcher


jolynch:
@hightoxicity Thanks!

One of these days we'll add specs for this darn class ...

hightoxicity:
Hi!
It is us which thank you for your sharing spirit! Your project is useful for us at Blablacar.

Tony

PR #153, opened 2015-12-23
Aurora support

jolynch:
This pulls in #148 with some of #104.

@JustinVenus, @benley does this work for both of you?

... we really need some specs for this class

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,25 +141,38 @@ Each value in the services hash is also a hash, and should contain the following
 We've included a number of `watchers` which provide service discovery.
 Put these into the `discovery` section of the service hash, with these options:
 
-##### Stub #####
+##### Base #####
 
-The stub watcher is useful in situations where you only want to use the servers in the `default_servers` list.
+The base watcher is useful in situations where you only want to use the servers in the `default_servers` list.
 It has only one option:
 
-* `method`: stub
+* `method`: base
 
 ##### Zookeeper #####
 
 This watcher retrieves a list of servers from zookeeper.
-It takes the following options:
+It takes the following mandatory arguments:
 
 * `method`: zookeeper
 * `path`: the zookeeper path where ephemeral nodes will be created for each available service server
 * `hosts`: the list of zookeeper servers to query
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -141,25 +141,38 @@ Each value in the services hash is also a hash, and should contain the following
...
+
+###### Decoding service nodes ######
+Synapse attempts to decode the data in each of these nodes using JSON and you can control how it is decoded with the `decode` argument. If provided, the `decode` hash should contain the following:
+
+* `method` (one of ['nerve', 'aurora'], default: 'nerve'): The kind of data to expect in zookeeper
+* `port_name` (default: nil): If using the 'aurora' method, this controls which of the `additionalEndpoints` is chosen.
+
+If the `method` is `nerve`, then we expect to find nerve registrations with a `host` and a `port`.
+
+If the `method` is `aurora` then we expect to find aurora registrations with a `serviceEndpoint` and optionally one or more `additionalEndpoints`.
```
benley: might be more appropriate to call this method "serverset", since it's actually the finagle serverset format that aurora happens to share.

benley:
Crap, I totally forgot about updating my branch!  Anyway yes, this looks like it will work.  Just see the one comment about naming.

JustinVenus:
lgtm #shipit

README.md
jolynch: Ok, from [the source](https://github.com/twitter/finagle/blob/master/finagle-serversets/src/main/scala/com/twitter/finagle/zookeeper/ZookeeperServerSetCluster.scala) it appears they select ports by "endpointName", so I'll do the same to be consistent.

jolynch:
@benley, @JustinVenus 
I've merged the branch (only change from initial PR was that we call it the `serverset` method instead of the `aurora` method and `port_name` became `endpoint_name`). When you guys get a chance to give it a try in your setups please let me know if this works for you two.

Some basic manual testing on my end seems to indicate it's ok, but we don't run it in production so I don't really want to claim that.

PR #160, opened 2016-02-01
Fix service discovery under Aurora

dfboyd256:
It logs "ERROR -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse:
invalid data in ZK node member_0000000004 at
/aurora/aa/ubuntu/devel/rising: uninitialized constant
Synapse::ServiceWatcher::ZookeeperWatcher::ValueError"

It does this because the host and port are inside another level of JSON
hash in an Aurora ServerSet record; they're not out at toplevel.

### DIFF
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -240,8 +240,8 @@ class Synapse::ServiceWatcher
       log.debug "synapse: deserializing process data"
       decoded = @decode_method.call(data)
 
-      host = decoded['host'] || (raise KeyError, 'instance json data does not have host key')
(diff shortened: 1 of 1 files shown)


jolynch:
@dfboyd256 How are you configuring your decode method? If you don't supply an endpoint name you should get [``serverset_decode``](https://github.com/airbnb/synapse/blob/00024028b3664ca8a34da1db4c8c00cc08c7614a/lib/synapse/service_watcher/zookeeper.rb#L96) which will return the data under serviceEndpoint.

jolynch:
I'm referring to [this configuration](https://github.com/airbnb/synapse#decoding-service-nodes)

dfboyd256:
I was having build problems; found out I was running an old version.

PR #162, opened 2016-02-10
Add a note about the state file's use for monitoring Synapse

hashbrowncipher:
Documenting this behavior could help other Synapse users.


jolynch:
Thanks :-)

PR #164, opened 2016-02-11
Service watcher supports filtering backends by label 

scarletmeow:
### Overview

The high level context for this commit: we want to allow users to configure Synapse to distinguish different pools of service instances and to route traffic between them based on user configuration.

Original Discussion PR: https://github.com/airbnb/nerve/pull/81

### Details

Synapse's `discovery` configuration will support a new section `label_filter`. Let's assume the user is interested in setting up failover from the local availability zone to other nodes in the same region and eventually failing over to any available global instance.

```json
{ 
  "example_service":  {
    "haproxy": {
      "backend_name": "example_service_default",
      "frontend": [
        "acl use_local_backend nbsrv(example_service_local) gt 2",
        "acl use_regional_backend nbsrv(example_service_regional) gt 2",
        "use_backend example_service_local use_local_backend",
        "use_backend example_service_regional use_regional_backend",
      ],
    },
  }
}
```

This configuration comes with additional service definitions with different `label_filter` options:

```json
{
  "example_service_local": {
    "discovery": {
      "label_filter": { "label": "az", "value": "us-east-1a", "condition": "equals" },
    }
},
  "example_service_regional": {
    "discovery": {
      "label_filter": { "label": "region", "value": "us-east-1", "condition": "equals" },
    }
  }
}
```

Given a set of service instances for `my-service` reporting the following service data:
```json
{
  "somehost:1111": { "host": "somehost", "port": 1111, "labels": { "az": "us-east-1a", "region": "us-east-1" } },
  "somehost:2222": { "host": "somehost", "port": 2222, "labels": { "az": "us-east-1a", "region": "us-east-1" } },
  "somehost:3333": { "host": "somehost", "port": 3333, "labels": { "az": "us-east-1c", "region": "us-east-1" } },
  "somehost:4444": { "host": "somehost", "port": 4444, "labels": { "az": "us-west-2a", "region": "us-west-2" } },
  "somehost:5555": { "host": "somehost", "port": 5555, "labels": { "az": "eu-central-1a", "region": "eu-central" } }
}
```

Synapse is expected to produce the following HAProxy configuration:
```
frontend example_service
  acl use_local_backend nbsrv(example_service_local) gt 2
  acl use_regional_backend nbsrv(example_service_regional) gt 2
  use_backend example_service_local use_local_backend
  use_backend example_service_regional use_regional_backend
  default_backend example_service_default

backend example_service_default
  server somehost:1111 somehost:1111 cookie somehost:1111
  server somehost:2222 somehost:2222 cookie somehost:2222
  server somehost:3333 somehost:3333 cookie somehost:3333
  server somehost:4444 somehost:4444 cookie somehost:4444
  server somehost:5555 somehost:5555 cookie somehost:5555

backend example_service_local
  server somehost:1111 somehost:1111 cookie somehost:1111
  server somehost:2222 somehost:2222 cookie somehost:2222

backend example_service_regional
  server somehost:1111 somehost:1111 cookie somehost:1111
  server somehost:2222 somehost:2222 cookie somehost:2222
  server somehost:3333 somehost:3333 cookie somehost:3333

```

### DIFF
--- a/lib/synapse/service_watcher/base.rb
+++ b/lib/synapse/service_watcher/base.rb
@@ -21,6 +21,7 @@ class Synapse::ServiceWatcher
 
       @name = opts['name']
       @discovery = opts['discovery']
+      @label_filter = @discovery['label_filter'] || false
 
       @leader_election = opts['leader_election'] || false
       @leader_last_warn = Time.now - LEADER_WARN_INTERVAL
@@ -86,6 +87,8 @@ class Synapse::ServiceWatcher
 
         # if leader election fails, return no backends
         return []
+      elsif @label_filter
+        return filter_backends_by_label(@backends, @label_filter)
       end
 
       return @backends
@@ -97,6 +100,17 @@ class Synapse::ServiceWatcher
         unless @discovery['method'] == 'base'
 
       log.warn "synapse: warning: a stub watcher with no default servers is pretty useless" if @default_servers.empty?
+    end
+
+    def filter_backends_by_label(backends, label_filter)
+      filtered_backends = []
+      backends.each do |backend|
+        backend_labels = backend['labels'] || {}
+        if label_filter['condition'] == 'equals' and backend_labels[label_filter['label']] == label_filter['value']
+          filtered_backends << backend
+        end
+      end
+      return filtered_backends
     end
 
     def set_backends(new_backends)
--- a/spec/lib/synapse/service_watcher_base_spec.rb
+++ b/spec/lib/synapse/service_watcher_base_spec.rb
@@ -109,5 +109,31 @@ describe Synapse::ServiceWatcher::BaseWatcher do
         expect(subject.backends).to eq(backends + default_servers)
       end
     end
+
+    context 'with label_filter set' do
+      let(:matching_labeled_backends) { [
+        { 'name' => 'server1', 'host' => 'server1', 'port' => 1111, 'labels' => { 'az' => 'us-east-1a' } },
+        { 'name' => 'server2', 'host' => 'server2', 'port' => 2222, 'labels' => { 'az' => 'us-east-1a' } },
+      ] }
+      let(:non_matching_labeled_backends) { [
+        { 'name' => 'server3', 'host' => 'server3', 'port' => 3333, 'labels' => { 'az' => 'us-west-1c' } },
+        { 'name' => 'server4', 'host' => 'server4', 'port' => 4444, 'labels' => { 'az' => 'us-west-2a' } },
+      ] }
+      let(:non_labeled_backends) { [
+        { 'name' => 'server5', 'host' => 'server5', 'port' => 5555 },
+      ] }
+      let(:args) {
+        testargs.merge({ 'discovery' => {
+          'method' => 'base',
+          'label_filter' => { 'condition' => 'equals', 'label' => 'az', 'value' => 'us-east-1a' } }
+        })
+      }
+      it 'removes all backends that do not match the label_filter' do
+        expect(subject).to receive(:'reconfigure!').exactly(:once)
+        subject.send(:set_backends, matching_labeled_backends + non_matching_labeled_backends +
+                     non_labeled_backends)
+        expect(subject.backends).to eq(matching_labeled_backends)
+      end
+    end
   end
 end
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -127,7 +127,7 @@ class Synapse::ServiceWatcher
 
         begin
           # TODO: Do less munging, or refactor out this processing
-          host, port, name, weight, haproxy_server_options = deserialize_service_instance(node.first)
+          host, port, name, weight, haproxy_server_options, labels = deserialize_service_instance(node.first)
         rescue StandardError => e
           log.error "synapse: invalid data in ZK node #{id} at #{@discovery['path']}: #{e}"
         else
@@ -141,7 +141,8 @@ class Synapse::ServiceWatcher
           new_backends << {
             'name' => name, 'host' => host, 'port' => server_port,
             'id' => numeric_id, 'weight' => weight,
-            'haproxy_server_options' => haproxy_server_options
+            'haproxy_server_options' => haproxy_server_options,
+            'labels' => labels
           }
         end
       end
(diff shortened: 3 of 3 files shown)


scarletmeow:
@SonicWang i put up an initial patchset, feedback is greatly appreciated!

igor47:
i was envisioning this configuration living higher (lower?) on the stack. i imagined us having labels as a top-level key for a service configuration (at the same level as discovery/haproxy). 

labels could work like leader election, here: https://github.com/airbnb/synapse/blob/master/lib/synapse/service_watcher/base.rb#L74

that is, the service watcher will not return any backends that do not match the required label configuration. haproxy would not need to be modified -- it would still configure the stanzas with the backends it knows about, it would just potentially not know about all of the things that are registered.

this has both advantages and drawbacks. the advantage is that we are decoupled from haproxy. if someone wants to add an nginx load balancer to synapse, they would not have to re-implement the logic. the same goes for the state file mechanism which is already in-place.

the drawback is that this punts more complexity back to synapse configuration. in our chef config, where we now only create a single synapse section for each smartstack service, we would have to create multiple synapse sections for each smartstack service. this seems easy enough for us, since we are using chef to programmatically generate the synapse config anyway, but might be more difficult in other environments.

what do you think of my alternate proposal? does that make sense?

igor47:
the backends function could be something like: 
```ruby
def backends
  if @leader_election
    elect_leader_backend
  elsif @labels
    filter_backends_by_labels
  else
    @backends
  end
end
```

jolynch:
I also prefer the filter approach, although I'd probably put the configuration for it in the ``discovery`` section since the filters should affect all frontends (haproxy, file, [nginx], w.e.).

@igor47 I'm not aware of any shop that would care about this kind of thing that doesn't automatically generate the configuration. You guys use chef, we use cron jobs, I've heard of folks using consul-template to generate synapse configs, etc ... As such, I think your proposal makes the right tradeoffs.

scarletmeow:
@igor47 Oh, that's interesting to hear that there are plans to extend this to beyond haproxy. If that's the case, then it makes sense to try to pull this functionality up a level.

I'm a little concerned that by making the different label filters separate services, there won't be a natural place to put the configuration describing how to fail over/route traffic to these labels. It also doesn't look like the ZK connection/discovery mechanism is pooled between the different services- each filter would be reconstructing their view of the current state. I suspect there are cases where the generated haproxy.cfg would be inconsistent (e.g. a service instance could potentially appear in more than one filter). Would this be a major issue? 

I think the advantage of having a single `labels` section is that it's a lot easier for users to reason about the behavior of filtering and it ensures the instances are always consistently partitioned. Having a default backend pool that will catch all unmatched instances is also a good backup in case service label values are corrupted somehow (for people who are using more dynamic values for label filters than AZ).

What do you think of changing the signature of  `backends` to return a hash of `{ backend_name: backends }` instead of a list of backends? (This also means modifying the behavior in `haproxy.rb` to expect service watchers to return more than one set of backends.)

This way configuration would naturally fit into the discovery section, similar to what @jolynch is proposing, except the configuration describes a switch statement that partitions the fetched backends. So we're still pulling up the configuration out from the `haproxy` section and into the `discovery` section, but we don't end up having the duplication of service configs.

scarletmeow:
Just spoke with Igor and Sonic about this. There's no requirement of strictly partitioning the nodes into the clusters (so it's OK if a node in us-east1a appears in the local, regional and global backends). The main consideration is that the synapse configuration should be easily constructed by chef (and by extension, any other provisioning system).

As for the DSL to describe label conditions, keeping it extremely simple for now would be sufficient (only support equality). I'll leave it up to future PRs to consider extending support for negation, multiple conditions, etc.

jolynch:
I dig it.

++ keeping the logic about datacenter topology in the configuration
++ keeping the label matching logic simple, adding complexity as people need it 
++ doing the hard work in the config generators (haproxy)

scarletmeow:
@igor47 The patchset is now amended as per your suggestion. Let me know what you think!

```json
{ "discovery": {
"method": "base",
 "label_filter": { "condition": "equals", "label": "az", "value": "us-east-1a" } }
```

`label_filter` can be extended by implementing different kinds of conditions, including compound conditions like `'or'` and `'and'`. (At that point, the actual filter evaluation logic should be its own helper function).

igor47:
oh, i like the approach. it meets our needs right now, and we can evolve by adding new `condition`s.

spec/lib/synapse/service_watcher_base_spec.rb
```diff
@@ -109,5 +109,26 @@ def remove_arg(name)
         expect(subject.backends).to eq(backends + default_servers)
       end
     end
+
+    context 'with label_filter set' do
+      labeled_backends = [
```
igor47: this should be initialized in a `let`.

also, for more clarity, might be nice to do something like this:
```ruby
let(:matching_labled_backends) { [] }
let(:non_matching_labled_backends) { [] }
let(:non_labled_backends) { [] }
```

then you can do
```ruby
subject.send(:set_backends, matching_labled_backends + non_matching + non_labled)
expect(subject.backends).to contain_exactly(matching_labled_backends)
```

igor47:
lgtm, i suggested an improvement for the tests

spec/lib/synapse/service_watcher_base_spec.rb
scarletmeow: That's a great suggestion. Done!

scarletmeow:
@jolynch @SonicWang let me know if this is good to merge!

SonicWang:
:+1: LGTM!

scarletmeow:
Updated title and description to match commit.

lib/synapse/service_watcher/base.rb
```diff
@@ -99,6 +102,17 @@ def validate_discovery_opts
       log.warn "synapse: warning: a stub watcher with no default servers is pretty useless" if @default_servers.empty?
     end
 
+    def filter_backends_by_label(backends, label_filter)
+      filtered_backends = []
+      backends.each do |backend|
+        backend_labels = backend['labels'] || {}
+        if label_filter['condition'] == 'equals' and backend_labels[label_filter['label']] == label_filter['value']
```
igor47: oops, i should have caught this earlier. using `and` is not a good idea in ruby, you should always use `&&` because `and` has odd precedence rules. i'm working on this section of code, so i'll just fix it in my PR.

PR #165, opened 2016-02-16
Permit to hash server identifier for backend affinity stored in cookie

hightoxicity:
Avoid to send server + port in clear to endusers

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,7 @@ This section is its own hash, which should contain the following keys:
   (defaults to the service's key in the `services` section)
 * `listen`: these lines will be parsed and placed in the correct `frontend`/`backend` section as applicable; you can put lines which are the same for the frontend and backend here.
 * `shared_frontend`: optional: haproxy configuration directives for a shared http frontend (see below)
+* `cookie_hash_servernames`: optional: if true, synapse hashes backend names on cookie value assignation of your discovered backends, useful in http mode and when you want to use haproxy cookie feature but you do not want that your end users receive a Set-Cookie with your server name and ip readable in clear
 
 <a name="haproxy"/>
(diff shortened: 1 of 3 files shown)


jolynch:
Hi @hightoxicity, thanks for the PR.

Can you give some context for why this is useful / what problem you're trying to resolve?

I think I understand why this can't be in server_options, so I believe this complexity makes sense in Synapse instead of the configuration layer. If we're going to merge this though we need a test (see `spec/lib/synapse/haproxy_spec.rb`) and we need to update the README.

hightoxicity:
Hi @jolynch, great thanks for your reply,

I think this is a good thing to set a cookie value for discovered backends (in http mode, it allows us to make some affinity over synapse configuration). I am not searching to implement cookie persistence but only backend affinity.
I am currently using this feature : https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#4-cookie
I do not want to keep affinity state on haproxy server using hash maps, I know it is possible but I prefer to let the user keep the information (using _cookie PHPSESSID prefix_), aim is to do some level 7 throttling easily behind..
The fact is that without any hash my endusers can see in clear some critical information about our network topology through their cookies (ip + port of the proxified backend).
Since the name must persist (be stable) to accomplish an affinity over several synapse nodes, I propose to use a sha1 hash on discovered backend name to be able to use haproxy cookie feature and masking ip + port in served cookies.

I added a test.
I do not know if my use case is clear.

Great thanks.
Regards.

Tony.

hightoxicity:
Hi guy I need to submit another PR, so I close this one ,create a branch and submit a new PR for that.
Sorry I am just a newbie in github contibuting.

Tony.

PR #166, opened 2016-02-16
Add `backend_order` option support to `haproxy`

SonicWang:
Currently we are doing shuffling (random ordering) only because we wanted to load
balance with round-robin to be more even across all machines.

This option allows one to specify how backends are ordered
in the final backend stanza configuration. This make it possible to support
usages that require an identical backend stanza configuration across machines.

@igor47  @jolynch  

cc @scarletmeow

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,7 @@ This section is its own hash, which should contain the following keys:
 * `backend_name`: The name of the generated HAProxy backend for this service
   (defaults to the service's key in the `services` section)
 * `listen`: these lines will be parsed and placed in the correct `frontend`/`backend` section as applicable; you can put lines which are the same for the frontend and backend here.
+* `backend_order`: optional: how backends should be ordered in the `backend` stanza. (default is shuffling). Setting to `asc` means sorting backends in ascending alphabetical order before generating stanza. `desc` means descending alphabetical order. `no_shuffle` means no shuffling or sorting.
 * `shared_frontend`: optional: haproxy configuration directives for a shared http frontend (see below)
 
 <a name="haproxy"/>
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -259,6 +259,7 @@ This section is its own hash, which should contain the following keys:
 * `backend_name`: The name of the generated HAProxy backend for this service
   (defaults to the service's key in the `services` section)
 * `listen`: these lines will be parsed and placed in the correct `frontend`/`backend` section as applicable; you can put lines which are the same for the frontend and backend here.
+* `backend_order`: optional: how backends should be ordered in the `backend` stanza. (default is shuffling). Setting to `asc` means sorting backends in ascending alphabetical order before generating stanza. `desc` means descending alphabetical order. `no_shuffle` means no shuffling or soring.
```
scarletmeow: nit: should be `shuffling or sorting`

scarletmeow:
Maybe an unexpected side effect, but I noticed some of the names are formed by `hostname:port`, so and ascending sort will not sort things lexicographically rather than following the port number.

There's a key that services expose today in order to elect a leader (if configured), maybe that key can be re-used for sort order? See https://github.com/airbnb/synapse/blob/master/lib/synapse/service_watcher/base.rb#L77

SonicWang:
A kind ping @igor47  @jolynch. I wonder if you guys have any comments on exposing this config as the way the PR does?

jolynch:
:+1: 

This lg2m, and closes #127 for real.

@asiragusa you might be interested in this.

SonicWang:
Thanks for taking a look @jolynch !

asiragusa:
Thanks guys!

PR #168, opened 2016-02-18
zookeeper_dns is broken

scarletmeow:
While testing the new release of synapse v0.13.0, I noticed that the zookeeper_dns discovery method was failing with the following error:

```
E, [2016-02-17T22:03:58.722362 #10392] ERROR -- 
Synapse::ServiceWatcher::ZookeeperDnsWatcher::Zookeeper: synapse: invalid data in ZK node 
example-service at /production/services/example-service/services: undefined method `call' for 
nil:NilClass
```

The root cause is that `@decode_method` is not properly initialized for the class, since it is initialized in `validate_discovery_opts` which is overwritten in ZookeeperDNSWatcher::Zookeeper. We can't call `super` since it's valid for the discovery method to be `zookeeper`.

There's probably a more elegant fix, let me think about it. The other issue is if we should yank v0.13.0 of synapse since it is broken for this discovery method.

@SonicWang @igor47 @jolynch

### DIFF
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -11,6 +11,24 @@ class Synapse::ServiceWatcher
     @@zk_pool_count = {}
     @@zk_pool_lock = Mutex.new
 
+    def initialize(opts={}, synapse)
+      super(opts, synapse)
+
+      # Alternative deserialization support. By default we use nerve
+      # deserialization, but we also support serverset registries
+      @decode_method = self.method(:nerve_decode)
+      if @discovery['decode']
+        valid_methods = ['nerve', 'serverset']
+        decode_method = @discovery['decode']['method']
+        unless decode_method && valid_methods.include?(decode_method)
+          raise ArgumentError, "missing or invalid decode method #{decode_method}"
+        end
+        if decode_method == 'serverset'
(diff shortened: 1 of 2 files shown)


scarletmeow:
A less hacky fix could be to change the hardcoded `'zookeeper'` value to a class value that is overwritten to zookeeper_dns. This would mean `validate_discovery_opts` in the subclass ZookeeperDNS can call the super method directly, killing the existing code duplication. https://github.com/airbnb/synapse/pull/168/files#diff-86201d0784bc64799b300dce40bd1d45L39

SonicWang:
Fundamentally  `validate_discovery_opts` should not change any state. That's the thing we should fix. I think on top of your current change you can just move https://github.com/airbnb/synapse/blob/master/lib/synapse/service_watcher/zookeeper.rb#L47-L57 to the initialize method. This way both `zookeeper` and `zookeeper_dns` will be able to parse either nerve or serverset formats.

scarletmeow:
@SonicWang @igor47 I moved the block of logic that sets the deserialization setting up into the `initialize` function.

SonicWang:
:speedboat:

jolynch:
:+1:

jolynch:
Thanks for fixing this, oops...

PR #170, opened 2016-02-18
Be able to use http-reuse to make connections pooling with backends

hightoxicity:
Be able to use http-reuse to make connections pooling with backends (introduced with hap 1.6)


jolynch:
Thanks!

Yea, we're going to need to sync up with 1.6 at some point, it's been a while ...

hightoxicity:
:-) Thank you guy !

PR #171, opened 2016-02-18
Permit to hash servernames in affinity cookie

hightoxicity:
Hi, this PR replace : https://github.com/airbnb/synapse/pull/165

Sorry again.

Tony

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,7 @@ This section is its own hash, which should contain the following keys:
 * `listen`: these lines will be parsed and placed in the correct `frontend`/`backend` section as applicable; you can put lines which are the same for the frontend and backend here.
 * `backend_order`: optional: how backends should be ordered in the `backend` stanza. (default is shuffling). Setting to `asc` means sorting backends in ascending alphabetical order before generating stanza. `desc` means descending alphabetical order. `no_shuffle` means no shuffling or sorting.
 * `shared_frontend`: optional: haproxy configuration directives for a shared http frontend (see below)
+* `cookie_value_method`: optional: default value is `name`, it defines the way your backends receive a cookie value in http mode. If equal to `hash`, synapse hashes backend names on cookie value assignation of your discovered backends, useful when you want to use haproxy cookie feature but you do not want that your end users receive a Set-Cookie with your server name and ip readable in clear.
 
 <a name="haproxy"/>
 ### Configuring HAProxy ###
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -1,6 +1,7 @@
 require 'fileutils'
 require 'json'
 require 'socket'
+require 'digest/sha1'
 
 module Synapse
   class Haproxy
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -261,6 +261,7 @@ This section is its own hash, which should contain the following keys:
 * `listen`: these lines will be parsed and placed in the correct `frontend`/`backend` section as applicable; you can put lines which are the same for the frontend and backend here.
 * `backend_order`: optional: how backends should be ordered in the `backend` stanza. (default is shuffling). Setting to `asc` means sorting backends in ascending alphabetical order before generating stanza. `desc` means descending alphabetical order. `no_shuffle` means no shuffling or sorting.
 * `shared_frontend`: optional: haproxy configuration directives for a shared http frontend (see below)
+* `cookie_hash_servernames`: optional: if true, synapse hashes backend names on cookie value assignation of your discovered backends, useful in http mode and when you want to use haproxy cookie feature but you do not want that your end users receive a Set-Cookie with your server name and ip readable in clear
```
jolynch: What if we called this `cookie_value_method` or some such and had it take values `'name'` (the default) or `'hash'`? That way if we need to support additional cookie values that can't be supplied in server_options we can.

jolynch:
@hightoxicity Just the one suggestion, but other than that this seems fine to me.

We probably need to factor out some of the backend stanza logic soon into separate functions since it's getting pretty complicated, but I think that's best for another time (and tests make it easier).

hightoxicity:
Hi @jolynch,

I have read your solution proposal with the cookie_value_method, it appears to be more evolutive, so I have try to make the implementation, I let you check.

Thanks for your help.

Tony.

lib/synapse/haproxy.rb
```diff
@@ -1,12 +1,18 @@
...
 require 'json'
 require 'socket'
+require 'digest/sha1'
 
 module Synapse
   class Haproxy
     include Logging
     attr_reader :opts, :name
 
+    @@cookie_value_methods = [
```
jolynch: Let's not actually validate this. Similar to `backend_order` option let's just have a case statement with an else that provides the default behavior.

lib/synapse/haproxy.rb
```diff
@@ -734,13 +740,28 @@ def generate_backend_stanza(watcher, config)
       else
         backends.keys.shuffle
       end
+
+      cookie_value_method = "name"
```
jolynch: This seems redundant to the block down below. Let's just do the case down below (I'll comment with a suggestion)

lib/synapse/haproxy.rb
```diff
@@ -734,13 +740,28 @@ def generate_backend_stanza(watcher, config)
...
+      end
+
       stanza = [
         "\nbackend #{watcher.haproxy.fetch('backend_name', watcher.name)}",
         config.map {|c| "\t#{c}"},
         keys.map {|backend_name|
           backend = backends[backend_name]
           b = "\tserver #{backend_name} #{backend['host']}:#{backend['port']}"
-          b = "#{b} cookie #{backend_name}" unless config.include?('mode tcp')
+          unless config.include?('mode tcp')
```
jolynch: How about something like:
```ruby
unless config.include?('mode tcp')
  b = case watcher.haproxy['cookie_value_method']
  when 'hash'
    b = "#{b} cookie #{Digest::SHA1.hexdigest(backend_name)}"
  else
    b = "#{b} cookie #{backend_name}"
  end
end
```

jolynch:
@hightoxicity A couple of suggestions :-)

hightoxicity:
Hi @jolynch, I have done like you proposed.

Thank you guy!
Tony

igor47:
thanks for the PR, looks great!

hightoxicity:
Hi @jolynch, I hope you are fine. I permited to try to enter in contact with you on linkedin (Tony Fouchard), we are about to do a cultural trip in San Francisco (a team of 6 Blablacar employes) next week before Velocity Conference in Santa Clara. Do you think we could meet us at Yelp next week? Sorry to write you here and many thanks for your attention. We could exchange on things like how we manage frontend/backend performance, we have also people working and interested in payment systems. Thanks.

Issue #172, opened 2016-02-18
DnsWatcher < BaseWatcher


bkochendorfer:
We were trying out this service at our organization in an internal datacenter and we temporarily lost internet connection. Because this is resolving the external address of [airbnb](https://github.com/airbnb/synapse/blob/01a2409a5f1e19742ff9380ce683a54b31a4b7ad/lib/synapse/service_watcher/dns.rb#L18) being required as part of 'ping?' it caused synapse to fail entirely. Perhaps throwing a recoverable error rather than a runtime exception would help prevent the entire cluster from collapsing.

```bash
I, [2016-02-17T17:34:09.333142 #8263]  INFO -- Synapse::Haproxy: synapse: restarted haproxy
E, [2016-02-18T16:58:00.932196 #8263] ERROR -- Synapse::Synapse: synapse: encountered unexpected exception #<RuntimeError: synapse: service watcher mongo failed ping!> in main thread
W, [2016-02-18T16:58:00.932830 #8263]  WARN -- Synapse::Synapse: synapse: exiting; sending stop signal to all watchers
I, [2016-02-18T16:58:00.932962 #8263]  INFO -- Synapse::ZookeeperDnsWatcher: synapse: stopping watcher mongo using default stop handler
W, [2016-02-18T16:58:00.933561 #8263]  WARN -- Synapse::ZookeeperDnsWatcher::Zookeeper: synapse: zookeeper watcher exiting
I, [2016-02-18T16:58:00.934565 #8263]  INFO -- Synapse::ZookeeperDnsWatcher::Zookeeper: synapse: zookeeper watcher cleaning up
I, [2016-02-18T16:58:00.934961 #8263]  INFO -- Synapse::ZookeeperDnsWatcher::Zookeeper: synapse: closing zk connection to 172.16.151.82:2181,172.16.151.86:2181,172.16.151.87:2181
I, [2016-02-18T16:58:00.937347 #8263]  INFO -- Synapse::ZookeeperDnsWatcher::Zookeeper: synapse: zookeeper watcher cleaned up successfully
[...]

PR #173, opened 2016-02-19
[issue-172:dns] Gracefully fail DNS resolver interruptions

JustinVenus:
Relates to https://github.com/airbnb/synapse/issues/172.

Synapse crashes if it can't resolve airbnb.com,  the specific exception not handled is RuntimeError.  I personally don't care that the resolver check is hard-coded to airbnb.com.  I also get that synapse is designed to be fail fast.  However this fail fast approach has an interesting side effect in certain use cases. 

We were testing a pre-production Mesos cluster with synapse as our service discovery layer.  We temporarily lost the ability to resolve DNS and synapse crashed on resolution failure.  Normally this isn't a big deal.  

We are also using Apache Aurora (aka TwitterScheduler) for Mesos and we are running synapse as a process member of the job's task.  Under Aurora/Thermos if a single process exits non-zero the whole task fails.  Normally this isn't a big deal either.

However since DNS resolution was interrupted, and synapse bailed on resolving 'airbnb.com' (every synapse process through the fleet failed within seconds of each other.  This caused every Task that synapse (using zookeeper_dns) was part of to fail.  This in turn caused every instance of each Job to fail.  This of course caused the jobs to be rescheduled.  Had this been a production sized cluster, we would have crushed the storage layer of our stack.

### DIFF
--- a/lib/synapse/service_watcher/dns.rb
+++ b/lib/synapse/service_watcher/dns.rb
@@ -15,7 +15,13 @@ class Synapse::ServiceWatcher
     end
 
     def ping?
-      @watcher.alive? && !(resolver.getaddresses('airbnb.com').empty?)
+      begin
+        @watcher.alive? && !(resolver.getaddresses('airbnb.com').empty?)
+      rescue => e
+        log.warn "Name server error: #{e.inspect}"
+        log.warn e.backtrace
+        false
+      end
     end
 
     def discovery_servers


jolynch:
As written this won't work because the [main loop ping check](https://github.com/airbnb/synapse/blob/master/lib/synapse.rb#L52-L55). Synapse really is is designed for the most part to be crash-fail instead of crash-recovery (with the exception of running the HAProxy reload command which is crash-recovery). Partly this is because it's easier to implement, and partly because it encourages good practice in making sure your entire control plane (Synapse/Zookeeper/w.e.) can crash and your data plane (HAProxy) keeps on chugging.

I changed Nerve to crash-recover in [our nerve fork](https://github.com/Yelp/nerve/pull/4/files) prior to merging it back upstream. I'd be ok with a similar strategy for Synapse (detecting that Watchers had failed and trying to re-establish them). @igor47 what are your thoughts on a change like that?

JustinVenus:
@jolynch I see, I missed the failed ping as the origin of the exception.

I do like the way your nerve fork recovers and that's essentially what I'm looking for in Synapse.

jolynch:
Just to clarify something, that is how airbnb/nerve works (I merged our fork back into upstream).

I'm on-board for a similar piece of functionality for Synapse, basically if something fails a ping reap the watcher and relaunch. The only worry there would be if you had a flakey watcher you might end up restarting HAProxy a lot (check out `restart_interval` and `restart_jitter` in the config options for how to prevent that). 

Actually it would be _more_ reliable if Synapse auto-remediated watchers because as it is right now every time Synapse starts we restart HAproxy regardless of how recently it was reloaded. If we keep the process up we can make use of the `restart_interval` parameter to properly limit HAProxy reloads.

igor47:
@jolynch fair point about HAProxy restarts; i would be okay with an overhaul of synapse to make it recover, especially since we've already done that to nerve (and are going to do it more to nerve with your change to reload the config without restarting).

PR #174, opened 2016-02-19
Create an option to stop ignoring weights. By default weights will still be ignored. 

minkovich:
This commit is based on the following by @bobtfish 
https://github.com/airbnb/synapse/pull/131

but does the following things differently:
1. By default weights are ignored. This is to maintain current behavior for safety.
2. HAProxy will reconfigure if weights changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,7 @@ This section is its own hash, which should contain the following keys:
 * `listen`: these lines will be parsed and placed in the correct `frontend`/`backend` section as applicable; you can put lines which are the same for the frontend and backend here.
 * `backend_order`: optional: how backends should be ordered in the `backend` stanza. (default is shuffling). Setting to `asc` means sorting backends in ascending alphabetical order before generating stanza. `desc` means descending alphabetical order. `no_shuffle` means no shuffling or sorting.
 * `shared_frontend`: optional: haproxy configuration directives for a shared http frontend (see below)
+* `ignore_weights`: optional: stops haproxy backend 'weight' options being generated, even if the Nerve registrations contain this information. This will cause all backend servers to be treated equally by haproxy. This defaults to true so weights will *NOT* be used by default. 
 
 <a name="haproxy"/>
 ### Configuring HAProxy ###
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -535,6 +535,7 @@ module Synapse
       @opts['do_writes'] = true unless @opts.key?('do_writes')
       @opts['do_socket'] = true unless @opts.key?('do_socket')
       @opts['do_reloads'] = true unless @opts.key?('do_reloads')
+      @opts['ignore_weights'] = true unless @opts.key?('ignore_weights')
 
       # how to restart haproxy
       @restart_interval = @opts.fetch('restart_interval', 2).to_i
@@ -741,6 +742,10 @@ module Synapse
           backend = backends[backend_name]
           b = "\tserver #{backend_name} #{backend['host']}:#{backend['port']}"
           b = "#{b} cookie #{backend_name}" unless config.include?('mode tcp')
+          if !@opts['ignore_weights'] && backend.has_key?('weight')
+            weight = backend['weight'].to_i
+            b = "#{b} weight #{weight}"
+          end
           b = "#{b} #{watcher.haproxy['server_options']}" if watcher.haproxy['server_options']
           b = "#{b} #{backend['haproxy_server_options']}" if backend['haproxy_server_options']
           b = "#{b} disabled" unless backend['enabled']
--- a/lib/synapse/service_watcher/base.rb
+++ b/lib/synapse/service_watcher/base.rb
@@ -117,7 +117,7 @@ class Synapse::ServiceWatcher
       # Aggregate and deduplicate all potential backend service instances.
       new_backends = (new_backends + @default_servers) if @keep_default_servers
       new_backends = new_backends.uniq {|b|
-        [b['host'], b['port'], b.fetch('name', '')]
+        [b['host'], b['port'], b.fetch('name', ''), b.fetch('weight', 1)]
       }
 
       if new_backends.to_set == @backends.to_set
--- a/spec/lib/synapse/haproxy_spec.rb
+++ b/spec/lib/synapse/haproxy_spec.rb
@@ -5,13 +5,16 @@ class MockWatcher; end;
 describe Synapse::Haproxy do
   subject { Synapse::Haproxy.new(config['haproxy']) }
 
-  let(:mockwatcher) do
+  def createmockwatcher(backends)
     mockWatcher = double(Synapse::ServiceWatcher)
     allow(mockWatcher).to receive(:name).and_return('example_service')
-    backends = [{ 'host' => 'somehost', 'port' => 5555}]
     allow(mockWatcher).to receive(:backends).and_return(backends)
     allow(mockWatcher).to receive(:haproxy).and_return({'server_options' => "check inter 2000 rise 3 fall 2"})
     mockWatcher
+  end
+
+  let(:mockwatcher) do
+    createmockwatcher [{ 'host' => 'somehost', 'port' => '5555'}]
   end
 
   let(:mockwatcher_with_server_options) do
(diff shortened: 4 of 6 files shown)


lib/synapse/haproxy.rb
```diff
@@ -741,6 +742,10 @@ def generate_backend_stanza(watcher, config)
           backend = backends[backend_name]
           b = "\tserver #{backend_name} #{backend['host']}:#{backend['port']}"
           b = "#{b} cookie #{backend_name}" unless config.include?('mode tcp')
+          if !@opts['ignore_weights'] && backend.has_key?('weight')
+            weight = backend['weight'].to_i
+            b = "#{b} weight #{weight}"
```
jolynch: So I believe that HAProxy will use the last weight provided. If that's true I think this should probably slot this in between the HAProxy backend `server_options` and the server `haproxy_server_options` level.

README.md
```diff
@@ -261,6 +261,7 @@ This section is its own hash, which should contain the following keys:
 * `listen`: these lines will be parsed and placed in the correct `frontend`/`backend` section as applicable; you can put lines which are the same for the frontend and backend here.
 * `backend_order`: optional: how backends should be ordered in the `backend` stanza. (default is shuffling). Setting to `asc` means sorting backends in ascending alphabetical order before generating stanza. `desc` means descending alphabetical order. `no_shuffle` means no shuffling or sorting.
 * `shared_frontend`: optional: haproxy configuration directives for a shared http frontend (see below)
+* `ignore_weights`: optional: stops haproxy backend 'weight' options being generated, even if the Nerve registrations contain this information. This will cause all backend servers to be treated equally by haproxy. This defaults to true so weights will *NOT* be used by default.
```
jolynch: Do you have strong feelings about this being an option? I'd prefer for folks to just not register weights unless they want them reflected in HAProxy.

lib/synapse/service_watcher/base.rb
```diff
@@ -117,7 +117,7 @@ def set_backends(new_backends)
       # Aggregate and deduplicate all potential backend service instances.
       new_backends = (new_backends + @default_servers) if @keep_default_servers
       new_backends = new_backends.uniq {|b|
-        [b['host'], b['port'], b.fetch('name', '')]
+        [b['host'], b['port'], b.fetch('name', ''), b.fetch('weight', 1)]
```
jolynch: This is slightly surprising behavior to me because it means while folks are switching weights (i.e. old nerve is running with weight=10 and new nerve is running with weight=20) we'd get both backends. I think I prefer not uniqing on weight (so we'll pick one backend or the other). What do you think?

README.md
igor47: i am also not a big fan of this option

lib/synapse/haproxy.rb
igor47: or maybe we should bail of server options also contains weight? i don't have a strong feeling as to what the precedence order should be... at least printing a warning to logs when there are contradictory options seems like a good idea.

this is probably indicating that it's time to refactor this function, it's getting pretty big.

lib/synapse/service_watcher/base.rb
igor47: we should probably either unique on a strictly limited set of properties (just host, port, possibly name) or on all of the possible properties.

lib/synapse/haproxy.rb
jolynch: Yea, I noted a similar concern about complexity in #171. I'm not sure if these PRs are the right place for that though or if we should refactor it once they're merged.

I agree printing a warning is useful, but I don't think we should bail out as I can understand use cases where someone might want to set some combination as a fallback scheme. `weight` and `haproxy_server_options` come from the registration side and `server_options` come from the discovery side. I can understand a situation where one side might have information the other does not.

README.md
minkovich: This is mainly for safety so people can rollout a new version without worrying about breaking something. I'm also not a fan of this option but it's kind of necessary unless you roll this into a major release.

lib/synapse/service_watcher/base.rb
minkovich: @jolynch, you're right. Let me fix that and add another test.

lib/synapse/haproxy.rb
minkovich: @igor47 and @jolynch, let me move these lines and add a warning when a conflict is detected.

lib/synapse/haproxy.rb
```diff
@@ -742,6 +743,14 @@ def generate_backend_stanza(watcher, config)
           b = "\tserver #{backend_name} #{backend['host']}:#{backend['port']}"
           b = "#{b} cookie #{backend_name}" unless config.include?('mode tcp')
           b = "#{b} #{watcher.haproxy['server_options']}" if watcher.haproxy['server_options']
+          if !@opts['ignore_weights'] && backend.has_key?('weight')
+            # Check if server_options already contains weight, is so log a warning
+            if watcher.haproxy['server_options'].include? 'weight'
+              log.info "synapse: weight is defined by server_options and by nerve"
```
scarletmeow: log.warn (nit)

README.md
scarletmeow: `ignore_weights` doesn't seem very useful as a setting and it is confusing that the default is shipped in a disabled state. There isn't a similar disable switch for `haproxy_server_options` and that's way more powerful than the weights setting (rhetorical question: why not just set `haproxy_server_options` to `weight 10`?). 

If the main concern is a breaking API change, then I would suggest bumping to a new major release for the next version. What do you all think?

jolynch: @minkovich I'm fine with us cutting a new minor (aka major in 0.X parlance) version if that's the concern. I think if folks are registering a ``weight`` they should expect to have it reflected in the load balancers.

@scarletmeow Weight is separate for a few reasons. The first is because it was merged before I came up with ``haproxy_server_options`` back when the plan was to merge all the things as top level keys. The second is because it can be dynamically updated over the stats socket making it much more useful for dynamically balancing load while most options that occur in ``haproxy_server_options`` (e.g. healthcheck, backup status. etc ...) cannot. The third is because it is frontend agnostic, weight is a common concept in all load balancers that I'm aware of.

minkovich: @jolynch @igor47 @scarletmeow I'm going to wipe out the option to ignore weights. It's going to simplify the code a lot. I'd just bold it in the announcement when you cut the next release. :)

lib/synapse/haproxy.rb
```diff
@@ -742,6 +742,14 @@ def generate_backend_stanza(watcher, config)
           b = "\tserver #{backend_name} #{backend['host']}:#{backend['port']}"
           b = "#{b} cookie #{backend_name}" unless config.include?('mode tcp')
           b = "#{b} #{watcher.haproxy['server_options']}" if watcher.haproxy['server_options']
+          if backend.has_key?('weight')
+            # Check if server_options already contains weight, is so log a warning
+            if watcher.haproxy['server_options'].include? 'weight'
```
jolynch: I think the backend (server) provided ``haproxy_server_options`` data is a more concerning conflict than ``server_options`` (a weight in the ``server_options`` would be like a default for servers or something).

jolynch:
I'm not clear on how we know to restart without something like line 715?

minkovich:
@jolynch 

> I'm not clear on how we know to restart without something like line 715?
Doesn't the new unit test in spec/lib/synapse/service_watcher_base_spec.rb show that reconfigure will be called when weights change? Doesn't that imply that HAProxy will be reconfigured? Am I missing something?

lib/synapse/haproxy.rb
minkovich: @jolynch, that's a good point. To reduce the number of iteration for this PR, could you suggest what message you'd like to see when haproxy_server_options also includes weight?

jolynch:
@minkovich Sorry I should have been more clear. A watcher calling set_backends is necessary but not  sufficient to determine a restart is needed (I believe your unit tests test that set_backend/reconfigure are called). Restarting has to be done via notification (by setting to True) on the ``@restart_required`` variable after interrogating the local HAProxy to ensure that the state that Synapse has (via set_backends) matches what is in the _running_ HAProxy which may or may not be what is in the configuration file because of stats socket updates.

The typical update ordering is ``set_backends`` -> ``reconfigure`` -> ``haproxy.update_config`` -> ``haproxy.update_backends`` (which may or may not set ``@restart_required``). There are also a few short cuts where we just set ``@restart_required`` directly (like the ``haproxy_server_options`` handling code).

Basically Synapse is juggling three state stores and trying to keep them consistent. It's trying to watch the registry (zk), reflect that state internally (synapse) and then ensure that this state is consistent with the routing component (haproxy). I think your change reflects the changes from zk -> synapse, but I don't see how it reflects them from synapse to haproxy.

PR #175, opened 2016-02-20
Feat fix test failure on glob

hightoxicity:
Hi,

I have tried to fix a test which breaks on my system due to the order of files returned by glob.
I am running:

> Linux debian 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt20-1+deb8u2 (2016-01-02) x86_64 GNU/Linux

Ruby doc says that files order returned by glob is system dependent. It is problematic for just this test since this is the only one which return 2 elements.

It should be now fixed for all systems.
Regards.

Tony


jolynch:
Thanks :-)

hightoxicity:
Thanks to you Jolynch! Tony.

PR #179, opened 2016-03-28
Weights support in Synapse

scarletmeow:
Follow up to https://github.com/airbnb/synapse/pull/174.

Adds logic to signal HAProxy restart if `weight` has changed.

Added an additional warning if `weights` is redefined in `haproxy_server_options`. I would prefer not to implement behavior that strips out any user-defined `weight X` statement in HAProxy server options. It seems sufficient to warn about the unexpected behavior in either case.


lib/synapse/haproxy.rb
```diff
@@ -742,6 +747,17 @@ def generate_backend_stanza(watcher, config)
           b = "\tserver #{backend_name} #{backend['host']}:#{backend['port']}"
           b = "#{b} cookie #{backend_name}" unless config.include?('mode tcp')
           b = "#{b} #{watcher.haproxy['server_options']}" if watcher.haproxy['server_options']
+          if backend.has_key?('weight')
+            # Check if server_options/haproxy_server_options already contains weight, if so log a warning
+            if watcher.haproxy.fetch('server_options', '').include? 'weight'
+              log.warn "synapse: weight is defined by server_options and by nerve"
+            end
+            if backend['haproxy_server_options'] and backend['haproxy_server_options'].include? 'weight'
```
scarletmeow: unfortunately, due to how the parsing logic in ServiceWatcher works today, `haproxy_server_options` always exists as a key in backend and defaults to value `nil`

lib/synapse/haproxy.rb
```diff
@@ -742,6 +747,17 @@ def generate_backend_stanza(watcher, config)
...
+              log.warn "synapse: weight is defined by haproxy_server_options and by nerve"
```
jolynch: I'd mention that the haproxy_server_options weight will take precedence.

lib/synapse/haproxy.rb
```diff
@@ -742,6 +747,17 @@ def generate_backend_stanza(watcher, config)
           b = "\tserver #{backend_name} #{backend['host']}:#{backend['port']}"
           b = "#{b} cookie #{backend_name}" unless config.include?('mode tcp')
           b = "#{b} #{watcher.haproxy['server_options']}" if watcher.haproxy['server_options']
+          if backend.has_key?('weight')
+            # Check if server_options/haproxy_server_options already contains weight, if so log a warning
+            if watcher.haproxy.fetch('server_options', '').include? 'weight'
+              log.warn "synapse: weight is defined by server_options and by nerve"
```
jolynch: I'd mention that the nerve weight will take precedence.

jolynch:
👍 

Looks good to me, although it looks like travis flaked out?

jolynch:
Also are you going to follow up with live update capability via the stats socket, or do we want to keep punting on that?

scarletmeow:
Gonna keep punting on the live update part for now. I'll be happy being able to merge this PR and close out the outstanding PRs from the two previous attempts at this.

lib/synapse/haproxy.rb
scarletmeow: Done

lib/synapse/haproxy.rb
scarletmeow: Done

scarletmeow:
@jolynch still good to merge this?

lib/synapse/haproxy.rb
```diff
@@ -714,6 +714,11 @@ def generate_backend_stanza(watcher, config)
             log.info "synapse: restart required because haproxy_server_options changed for #{backend_name}"
             @restart_required = true
           end
+          if (old_backend.fetch('weight', "") !=
```
igor47: so, it looks like this section of code is only invoked if the state file is enabled. i think this needs a little re-thinking...

scarletmeow: We detect a restart as long as the output config file is different: https://github.com/airbnb/synapse/blob/master/lib/synapse/haproxy.rb#L844. So you're right to point out this block of code as being out-of-place. I'm writing some tests to see if we even need any of the logic here.

PR #183, opened 2016-04-08
Fix spec/support/minimum.conf.yaml

tomfitzhenry:
* 'services' is a hash, not a list
* 'services' needs a 'haproxy' key


jolynch:
Well that test file is all sorts of messed up. @tomfitzhenry thanks for the PR, I'll make a followup to cleanup listen and such.

Issue #182, opened 2016-04-08
AWS Auto Scaling Group watcher


tomfitzhenry:

llarsson:
While definitely nice, this can alternatively be achieved by tagging the ASG and propagating the tag to the instances (and then using the tag watcher).

PR #184, opened 2016-04-08
[WIP] ASG watcher

tomfitzhenry:
Early feedback welcome.

Code mostly copied from ec2tag watcher.

Issue #182

### DIFF
--- a/lib/synapse/service_watcher/asg.rb
+++ b/lib/synapse/service_watcher/asg.rb
@@ -1,0 +1,139 @@
+require 'synapse/service_watcher/base'
+require 'aws-sdk'
+
+class Synapse::ServiceWatcher
+  class AsgWatcher < BaseWatcher
+
+    attr_reader :check_interval
+
+    def start
+      region = @discovery['aws_region'] || ENV['AWS_REGION']
+      log.info "Connecting to EC2 region: #{region}"
+
+      @asg_client = AWS::AutoScaling::Client.new(
+          region:            region,
+          access_key_id:     @discovery['aws_access_key_id']     || ENV['AWS_ACCESS_KEY_ID'],
+          secret_access_key: @discovery['aws_secret_access_key'] || ENV['AWS_SECRET_ACCESS_KEY'] )
+
+      @ec2_client = AWS::EC2::Client.new(
+          region:            region,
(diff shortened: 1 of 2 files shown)


lib/synapse/service_watcher/asg.rb
```diff
@@ -0,0 +1,139 @@
...
+      @ec2_client = AWS::EC2::Client.new(
+          region:            region,
+          access_key_id:     @discovery['aws_access_key_id']     || ENV['AWS_ACCESS_KEY_ID'],
+          secret_access_key: @discovery['aws_secret_access_key'] || ENV['AWS_SECRET_ACCESS_KEY'] )
+
+      @check_interval = @discovery['check_interval'] || 15.0
+
+      log.info "synapse: asg watcher looking for instances in asg #{@discovery["asg_name"]}"
+
+      @watcher = Thread.new { watch }
```
jolynch: Just a heads up there is a flaw with how the ec2tag watcher and other "watch in a Thread" watchers are implemented. Because this returns immediately from the thread creation synapse might proceed to restart HAProxy before watch has actually been called. The zookeeper watcher gets around this because it uses a callback system and waits for the callback to finish once before returning.

I'm not saying you have to change it, but I just want you to be aware of the shortcoming. I'm super open to ideas for how to fix it.

lib/synapse/service_watcher/asg.rb
```diff
@@ -0,0 +1,139 @@
...
+        begin
+          start = Time.now
+          if set_backends(discover_instances)
+            log.info "synapse: asg watcher backends have changed."
+          end
+        rescue Exception => e
+          log.warn "synapse: error in asg watcher thread: #{e.inspect}"
+          log.warn e.backtrace
+        ensure
+          sleep_until_next_check(start)
```
jolynch: Polling is definitely ok as a V1 or for small infras, but what are your thoughts on subscribing to a SNS topic of [autoscaling notifications](http://docs.aws.amazon.com/AutoScaling/latest/DeveloperGuide/ASGettingNotifications.html). It could be something super simple like "if we get any SNS notification at all, rediscover backends"

jolynch:
Looks pretty good. I think if we want this to really scale we'll have to do Pub/Sub via SNS notifications, but I dig it as a v1

PR #188, opened 2016-05-26
label_filter support for zookeeper_dns

scarletmeow:
Extends `label_filter` functionality for `ZookeeperDnsWatcher`. This requires a change to `ZookeeperDnsWatcher` to forward the discovery options for label filtering and a change to `DnsWatcher` to pass a long any server `labels`.

Tested that label functionality is working for `ZookeeperDnsWatcher` with this patchset.

to: @SonicWang @igor47 @jolynch

### DIFF
--- a/lib/synapse/service_watcher/dns.rb
+++ b/lib/synapse/service_watcher/dns.rb
@@ -85,6 +85,7 @@ class Synapse::ServiceWatcher
             'host' => address,
             'port' => server['port'],
             'name' => server['name'],
+            'labels' => server['labels'],
           }
         end
       end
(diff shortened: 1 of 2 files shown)


jolynch:
:+1: 
Code seems fine, can you add label_filter to the relevant docs in the readme?

Do we want to add a spec test for the zk dns watcher similar to the zk spec?

scarletmeow:
Thanks jolynch! I'll work on README docs in a separate PR.

For a spec test, I took a look at it earlier. The class ZookeeperDnsWatcher needs a refactor to make it more easily testable. I was concerned that maintaining these specs would be more work than their absence.

PR #189, opened 2016-05-26
Add label_filter to README.md

scarletmeow:
Adding documentation on `label_filter` to Synapse.

to: @jolynch


jolynch:
:+1: thanks

PR #191, opened 2016-05-26
Haproxy tests

igor47:
This PR contains no functional changes to Synapse. Instead, it adds a bunch of tests for the Haproxy class. We refactor the Haproxy class slightly to make testing easier. The main goal was to cover the state file handling with tests, since we're about to start using the state file at Airbnb. No bugs were found!

Ideally, we would only test the public methods of Haproxy -- `initialize`, `tick`, and `update_config`. however, if I set all of the other methods to private, we fail lots of existing tests, and testing the deep functionality of Haproxy via the public interface seems very tedious. So, although I intended to make that change here, I decided to save it for some other time.

to: @jolynch @scarletmeow

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -64,6 +64,7 @@ GEM
     slop (3.6.0)
     spoon (0.0.4)
       ffi
+    timecop (0.8.1)
     webmock (1.22.6)
       addressable (>= 2.3.6)
       crack (>= 0.3.2)
@@ -82,7 +83,8 @@ DEPENDENCIES
   rake
   rspec (~> 3.1.0)
   synapse!
+  timecop
   webmock
 
 BUNDLED WITH
-   1.10.6
+   1.11.2
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -6,7 +6,7 @@ require 'digest/sha1'
 module Synapse
   class Haproxy
     include Logging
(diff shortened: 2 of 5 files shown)


igor47:
oh, i recommend reading this one commit at a time.

Gemfile.lock
```diff
@@ -62,15 +64,21 @@ GEM
     rspec-support (3.1.2)
     safe_yaml (1.0.4)
     slop (3.6.0)
+    slyphon-log4j (1.2.15)
```
scarletmeow: So weirdly enough when I run bundle install on your branch I don't pick up these changes- the only delta for me is `timecop (0.8.1)`.

 json is `json (1.8.3)` and not `json (1.8.3-java)`. What's up with that and `slyphon`?

scarletmeow:
The code looks great to me, and I'm happy that this piece of functionality now has thorough test coverage. Each commit contains only a single logical change, so it was easy to see what was being refactored/tested.

I only have a comment about the `Gemfile.lock` changes. Is it expected that all these dependencies should be changed?

igor47:
That was a result of me running the `bundle install` under ruby 2.2.3 instead of ruby 1.9.3. i re-did the install under ruby 1.9.3 and the unexpected change are gone.

scarletmeow:
👍

jolynch:
Glad we have coverage of this now and that no bugs were found :-D

PR #192, opened 2016-05-26
Update whitelist for HAProxy 1.5.18 and 1.6.5

scarletmeow:
Generated using https://gist.github.com/igor47/7322131 against official HAProxy documentation.

I first ran it against [HAProxy 1.5 documentation](http://www.haproxy.org/download/1.5/doc/configuration.txt) (v1.5.18, last modified 10-May-2016 15:44) and then reverted any changes that removed existing options.

I then applied additive changes from  [HAProxy 1.6 documentation](http://www.haproxy.org/download/1.6/doc/configuration.txt) (v1.5.18, last modified 10-May-2016 15:42).

Finally, the script doesn't add options with `no ` prefix support like `no log` or `no option ...`, so these are added now.

@igor47 @jolynch @SonicWang


igor47:
do you need a rebase against master?

what happens if you run the script against 1.6 documentation?

weren't you saying something about supporting `no option` syntax?

scarletmeow:
I think https://github.com/airbnb/synapse/pull/193 needs to be merged to fix master, otherwise we are running tests against the published version of Synapse (for which the latest 0.13.2 hasn't been released to RubyGems yet, my bad) instead of Synapse source code

When I run it against 1.6, I get more changes and some stuff going away (`appsession` was replaced in a non-backwards compatible way). I'll add that to the PR too.

Still looking at the `no option` piece.

scarletmeow:
Temporarily rebasing against previous master commit 6cf3c85

scarletmeow:
Cleaning up the git history. All commits are additive changes to the HAProxy whitelist and are backwards compatible.

PR #193, opened 2016-05-27
Bump Gemfile.lock to fix specs

scarletmeow:
Synapse version in Gemfile.lock needs to get bumped. My guess is this is some kind of interaction with Travis and `--deployment` being enabled because we have a Gemfile.lock in the repository. Anything other than the current version will trigger an error similar to `Could not find synapse-0.13.2 in any of the sources`

to: @igor47

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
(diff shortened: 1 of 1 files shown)


scarletmeow:
Merging this ahead to unbreak tests on master

PR #194, opened 2016-05-27
[haproxy] Put backend['name'] first in constructing backend name

SonicWang:
This is needed because we are sorting backends using constructed
full string names. Currently we put `host:port` in front of `name`.
`host` could be an `ip` and in this case controlling ordering is hard.

Test: unit tests.

@scarletmeow @igor47 @jolynch

### DIFF
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -874,7 +874,7 @@ module Synapse
     def construct_name(backend)
       name = "#{backend['host']}:#{backend['port']}"
       if backend['name'] && !backend['name'].empty?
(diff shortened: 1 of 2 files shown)


SonicWang:
CI is failing due to gem version. https://github.com/airbnb/synapse/pull/193 should fix it.

scarletmeow:
I fixed master, please rebase.

This is a change only to how HAProxy configuration is generated and doesn't affect the internal state maintained by Synapse. As far as I can tell, nobody should be directly relying on the value of backend names in HAProxy. This should be safe to merge.

scarletmeow:
👍

PR #196, opened 2016-06-08
Clean up HAProxy sockets after reading

brianwolfe:
HAProxy closes its connection to the socket as soon as it finishes writing. This PR explicitly closes the socket from the synapse side. Without an explicit close, cleanup done on object finalization during garbage collection. This can cause synapse to temporarily accumulate a large number of socket file descriptors.

On a synthetic test with a large number of backends, this patch reduces the maximum number of open file descriptors from 940 to 52.

There are minor logging changes associated with this patch:
* added logging of error if an exception is raised when closing the file.
* if reading from the socket fails when called from https://github.com/airbnb/synapse/compare/master...brianwolfe:wolfe-socket-fd-cleanup?expand=1#diff-b2f6400f899968d5a5084ee66bf90b21R1106, there will be 2 error messages raised instead of 1, one message for the failure to read, one message for the failed command. Previously, only the failure to read error would be logged.

For old ruby versions, this cleanup can avoid this issue: https://bugs.ruby-lang.org/issues/8080 where glibc FORTIFY_SOURCE chokes on more than 1024 file descriptors despite linux supporting more.

@igor47

### DIFF
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -1031,32 +1031,28 @@ module Synapse
     end
 
     def haproxy_exec(command)
-      info = nil
       s = UNIXSocket.new(@opts['socket_file_path'])
       s.write(command)
       info = s.read
-    rescue StandardError => e
-      log.warn "synapse: unhandled error reading stats socket: #{e.inspect}"
-      @restart_required = true
-      nil
     ensure
-      unless s.nil?
-        begin
-          s.close
-        rescue StandardError => e
-          log.warn "synapse: unhandled error closing stats socket: #{e.inspect}"
-        end
+      if s
+        s.close
       end
-      info
     end
 
     # tries to set active backends via haproxy's stats socket
     # because we can't add backends via the socket, we might still need to restart haproxy
     def update_backends(watchers)
       # first, get a list of existing servers for various backends
-      info = haproxy_exec("show stat\n")
-
-      return if info.nil?
+      begin
+        stat_command = "show stat\n"
+        info = haproxy_exec(stat_command)
+      rescue StandardError => e
+        log.warn "synapse: restart required because socket command #{stat_command} failed "\
+                 "with error #{e.inspect}"
+        @restart_required = true
+        return
+      end
 
       # parse the stats output to get current backends
       cur_backends = {}
@@ -1103,10 +1099,18 @@ module Synapse
           end
 
           # actually write the command to the socket
-          output = haproxy_exec(command)
-          unless output == "\n"
-            log.warn "synapse: socket command #{command} failed: #{output}"
+          begin
+            output = haproxy_exec(command)
+          rescue StandardError => e
+            log.warn "synapse: restart required because socket command #{command} failed with "\
+                     "error #{e.inspect}"
             @restart_required = true
+          else
+            unless output == "\n"
+              log.warn "synapse: restart required because socket command #{command} failed with "\
+                      "output #{output}"
+              @restart_required = true
+            end
           end
         end
       end
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -1035,9 +1035,7 @@ module Synapse
       s.write(command)
       info = s.read
     ensure
-      if s
-        s.close
-      end
+      s.close if s
     end
 
     # tries to set active backends via haproxy's stats socket
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -1030,19 +1030,33 @@ module Synapse
       ]
     end
 
+    def haproxy_exec(command)
+      info = nil
+      s = UNIXSocket.new(@opts['socket_file_path'])
+      s.write(command)
+      info = s.read
+    rescue StandardError => e
+      log.warn "synapse: unhandled error reading stats socket: #{e.inspect}"
+      @restart_required = true
+      nil
+    ensure
+      unless s.nil?
+        begin
+          s.close
+        rescue StandardError => e
+          log.warn "synapse: unhandled error closing stats socket: #{e.inspect}"
+        end
+      end
+      info
+    end
+
     # tries to set active backends via haproxy's stats socket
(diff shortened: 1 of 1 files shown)


brianwolfe:
correction: reduction was to a maximum 53 file descriptors, not 52.

lib/synapse/haproxy.rb
```diff
@@ -1030,19 +1030,33 @@ def generate_backend_stanza(watcher, config)
...
+      nil
+    ensure
+      unless s.nil?
+        begin
+          s.close
+        rescue StandardError => e
+          log.warn "synapse: unhandled error closing stats socket: #{e.inspect}"
+        end
+      end
+      info
```
igor47: if you're thinking that this will return `info` from the function since the `ensure` block is the last one that runs, that's incorrect. this does nothing, and can be omitted. example:
```
irb(main):001:0> a = begin
irb(main):002:1* 5
irb(main):003:1> ensure
irb(main):004:1* 'hi'
irb(main):005:1> end
=> 5
irb(main):006:0> a
=> 5
```

lib/synapse/haproxy.rb
```diff
@@ -1030,19 +1030,33 @@ def generate_backend_stanza(watcher, config)
       ]
     end
 
+    def haproxy_exec(command)
+      info = nil
```
igor47: i think you don't need to initialize this here; this seems like it was done just to have something for the `ensure` block to reference

lib/synapse/haproxy.rb
```diff
@@ -1030,19 +1030,33 @@ def generate_backend_stanza(watcher, config)
...
+      s = UNIXSocket.new(@opts['socket_file_path'])
+      s.write(command)
+      info = s.read
+    rescue StandardError => e
```
igor47: i'm not sure about this `rescue` clause. i think we should allow the caller to decide how to handle exceptions, even though it does make the code more verbose at the call site. i can definitely envision cases where i want to query haproxy via the stats socket but *not* want to restart haproxy if that fails

lib/synapse/haproxy.rb
```diff
@@ -1030,19 +1030,33 @@ def generate_backend_stanza(watcher, config)
...
+      log.warn "synapse: unhandled error reading stats socket: #{e.inspect}"
+      @restart_required = true
+      nil
+    ensure
+      unless s.nil?
```
schleyfox: `if s` might as well use ruby truthiness.

lib/synapse/haproxy.rb
```diff
@@ -1089,18 +1103,10 @@ def update_backends(watchers)
...
           # actually write the command to the socket
-          begin
-            s = UNIXSocket.new(@opts['socket_file_path'])
-            s.write(command)
-            output = s.read()
-          rescue StandardError => e
-            log.warn "synapse: unknown error writing to socket"
+          output = haproxy_exec(command)
+          unless output == "\n"
+            log.warn "synapse: socket command #{command} failed: #{output}"
```
igor47: although this is unrelated to the main goal of the PR, i would love to see these log lines follow the `synapse: restart required` format we use elsewhere in this function.

```ruby
log.warn "synapse: restart required because socket command #{command} failed with out #{output}"
```

lib/synapse/haproxy.rb
```diff
@@ -1030,19 +1030,33 @@ def generate_backend_stanza(watcher, config)
...
-        s.write("show stat\n")
-        info = s.read()
-      rescue StandardError => e
-        log.warn "synapse: unhandled error reading stats socket: #{e.inspect}"
-        @restart_required = true
-        return
-      end
+      info = haproxy_exec("show stat\n")
+
+      return if info.nil?
```
igor47: i was confused here, because we appear to just exit without setting `@restart_required`. i suppose the expectation is that we only get `nil` output if there was an error, but this is why i was uncertain about the broad `rescue` block in the `haproxy_exec` function.

igor47:
okay, done with pass, let me know when you're ready for another look!

lib/synapse/haproxy.rb
brianwolfe: Ambivalent about it. I would lean towards having a helper method that does not trigger the restart in that case to keep it grouped where possible, but I'll pull up + reduplicate the rescue code for now.

lib/synapse/haproxy.rb
```diff
@@ -1030,16 +1030,26 @@ def generate_backend_stanza(watcher, config)
       ]
     end
 
+    def haproxy_exec(command)
+      s = UNIXSocket.new(@opts['socket_file_path'])
+      s.write(command)
+      info = s.read
+    ensure
+      if s
+        s.close
```
igor47: tighter: `s.close if s`

lib/synapse/haproxy.rb
```diff
@@ -1030,16 +1030,26 @@ def generate_backend_stanza(watcher, config)
...
     # tries to set active backends via haproxy's stats socket
     # because we can't add backends via the socket, we might still need to restart haproxy
     def update_backends(watchers)
       # first, get a list of existing servers for various backends
       begin
-        s = UNIXSocket.new(@opts['socket_file_path'])
-        s.write("show stat\n")
-        info = s.read()
+        stat_command = "show stat\n"
+        info = haproxy_exec(stat_command)
```
igor47: maybe do `info = haproxy_exec(stat_command) || ""` ? that way, if info is something funky, we'll just restart below because we don't find any backends, rather than hitting an error at `info.split`

brianwolfe: shouldn't be necessary. .read should just return empty string on no backends anyway.

igor47:
lgtm! i'll make a follow-up pr to bump the synapse version

lib/synapse/haproxy.rb
brianwolfe: http://ruby-doc.org/core-1.9.3/IO.html#method-i-read

> At end of file, it returns nil or "" depend on length. ios.read() and ios.read(nil) returns "". ios.read(positive-integer) returns nil.

We are not passing in a length argument, so we should get "".

PR #199, opened 2016-06-15
Multiple stats sockets

igor47:
we are running into haproxy CPU bottlenecks on some boxes. to get around those, we would like to run haproxy on multiple CPUs via `nbproc`. this PR enables the simplest way to configure `nbproc`, where all backends are listening on all processes and share the load. here is an example synapse config snapshot which would result in the desired behavior:

```json
{
  
  "haproxy": {
    "socket_file_path": [
      "/var/haproxy/stats1.sock",
      "/var/haproxy/stats2.sock",
    ],
    "global": [
      "nbproc 2",
      "cpu-map 1 0",
      "stats socket /var/haproxy/stats1.sock group smartstack mode 660 level admin process 1",
      "cpu-map 2 1",
      "stats socket /var/haproxy/stats2.sock group smartstack mode 660 level admin process 2",
```

this creates two independent haproxy processes, each with their own socket. we also tell synapse that two socket file paths exist. synapse will attempt to independently configure all of the processes via the stats socket, and will resort to restarting haproxy if we cannot dynamically configure either of them.

to: @nwadams @jtai 
cc: @jolynch @scarletmeow

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,7 @@ The top level `haproxy` section of the config file has the following options:
 * `do_writes`: whether or not the config file will be written (default to `true`)
 * `do_reloads`: whether or not Synapse will reload HAProxy (default to `true`)
 * `do_socket`: whether or not Synapse will use the HAProxy socket commands to prevent reloads (default to `true`)
+* `socket_file_path`: where to find the haproxy stats socket. can be a list (if using `nbproc`)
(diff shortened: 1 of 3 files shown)


nwadams:
lgtm

scarletmeow:
👍

PR #202, opened 2016-06-27
Refactor label filter

igor47:
In this PR, we make the following changes:

* go from `label_filter` to `label_filters` -- this allows us to define multiple filters, which must ALL match
* add a new filter condition, for `not-equals`
* allow leader election to operate on filtered backends, thus making leader election and filters no longer mutually exclusive (the mutual exclusivity was not documented)

this covers our main use case -- utilize only backends in the current region and the current az (the default case), or utilize only backends in the current region and NOT the current az (in case of AZ failover).

as well, we refactor the specs for labels to use a factory to create the labels, vs creating hashes manually and being forced to specify a bunch of attributes we don't care about.

to: @jolynch @scarletmeow @nwadams 
cc: @jtai

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,16 +152,17 @@ The base watcher is useful in situations where you only want to use the servers 
 It has the following options:
 
 * `method`: base
-* `label_filter`: optional filter to be applied to discovered service nodes
+* `label_filters`: optional list of filters to be applied to discovered service nodes
 
 ###### Filtering service nodes ######
-Synapse can be configured to only return service nodes that match a `label_filter` predicate. If provided, the `label_filter` hash should contain the following:
 
-* `label`: The label for which the filter is applied
+Synapse can be configured to only return service nodes that match a `label_filters` predicate. If provided, `label_filters` should be an array of hashes which contain the following:
+
+* `label`: The name of the label for which the filter is applied
 * `value`: The comparison value
-* `condition` (one of ['`equals`']): The type of filter condition to be applied. Only `equals` is supported at present
+* `condition` (one of ['`equals`', '`not-equals`']): The type of filter condition to be applied.
 
-Given a `label_filter`: `{ "label": "cluster", "value": "dev", "condition": "equals" }`, this will return only service nodes that contain the label value `{ "cluster": "dev" }`.
+Given a `label_filters`: `[{ "label": "cluster", "value": "dev", "condition": "equals" }]`, this will return only service nodes that contain the label value `{ "cluster": "dev" }`.
 
 ##### Zookeeper #####
 
--- a/lib/synapse/service_watcher/base.rb
+++ b/lib/synapse/service_watcher/base.rb
@@ -21,7 +21,14 @@ class Synapse::ServiceWatcher
 
       @name = opts['name']
       @discovery = opts['discovery']
-      @label_filter = @discovery['label_filter'] || false
+
+      # deprecated singular filter
+      @singular_label_filter = @discovery['label_filter']
+      unless @singular_label_filter.nil?
+        log.warn "synapse: `label_filter` parameter is deprecated; use `label_filters` -- an array"
+      end
+
+      @label_filters = [@singular_label_filter, @discovery['label_filters']].flatten.compact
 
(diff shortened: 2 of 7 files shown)


spec/factories/backend.rb
```diff
@@ -0,0 +1,20 @@
+FactoryGirl.define do
```
scarletmeow: Nice!

README.md
```diff
@@ -152,16 +152,17 @@ The base watcher is useful in situations where you only want to use the servers
...
-* `label`: The label for which the filter is applied
+Synapse can be configured to only return service nodes that match a `label_filters` predicate. If provided, `label_filters` should be an array of hashes which contain the following:
+
+* `label`: The name of the label for which the filter is applied
 * `value`: The comparison value
-* `condition` (one of ['`equals`']): The type of filter condition to be applied. Only `equals` is supported at present
+* `condition` (one of ['`equals`', '`not-equals`']): The type of filter condition to be applied.
 
-Given a `label_filter`: `{ "label": "cluster", "value": "dev", "condition": "equals" }`, this will return only service nodes that contain the label value `{ "cluster": "dev" }`.
+Given a `label_filters`: `[{ "label": "cluster", "value": "dev", "condition": "equals" }]`, this will return only service nodes that contain the label value `{ "cluster": "dev" }`.
```
scarletmeow: Should we warn that there's no validation to guard against nonsensical conditions? (ie, `az equals foo`, `az not-equals foo`)?

lib/synapse/service_watcher/base.rb
```diff
@@ -102,15 +120,16 @@ def validate_discovery_opts
...
-      backends.each do |backend|
+    def backends_filtered_by_labels
+      filtered_backends = @backends.select do |backend|
         backend_labels = backend['labels'] || {}
-        if label_filter['condition'] == 'equals' and backend_labels[label_filter['label']] == label_filter['value']
-          filtered_backends << backend
+        @label_filters.all? do |label_filter|
+          (label_filter['condition'] == 'equals' &&
+            backend_labels[label_filter['label']] == label_filter['value']) ||
+          (label_filter['condition'] == 'not-equals' &&
```
scarletmeow: Food for thought: So right now, if a node is missing a label, it makes sense for the `equals` label condition to fail (we're missing information, so it can't be `equal` unless we are expecting a `nil` or missing value). So there's no confusion here for the `equals` filter.

For `not-equals`, this scenario may cause confusion. When a backend node is missing the label value altogether, it will always pass the `not-equals` label condition. Right now, there's no chance of the labels information in Nerve being flaky as they're statically configured. This could be a concern in the future when label information is dynamic and labels could drop in/out of existence.

scarletmeow:
LGTM, thank you for cleaning up the code!

lib/synapse/service_watcher/base.rb
igor47: hmmm... good point. i wonder if i should explicitly check whether `backend_labels.include?(label_filter['label'])` before matching anything?

scarletmeow: As discussed offline:
- `not-equals` will give the opposite truth table for `equals` operator
- if users want to define behavior around missing or empty values, they can explicitly match against `nil`.

nwadams:
lgtm

README.md
```diff
@@ -152,16 +152,17 @@ The base watcher is useful in situations where you only want to use the servers
 It has the following options:
 
 * `method`: base
-* `label_filter`: optional filter to be applied to discovered service nodes
+* `label_filters`: optional list of filters to be applied to discovered service nodes
```
jolynch: What do you think about leaving label_filter in there for backwards compatibility. I know technically we're in 0.X releases here, but it's pretty easy to maintain backwards compatibility here (just accept both, append `label_filter` to `label_filters`.

jolynch:
👍 lg2m

I am curious about how we're going to allow users to combine the label filters going forward. In this iteration it's a logical and over all the filters, but I am curious how much complexity other options would be (such as or, negation, etc ...). For this PR an and sg2m.

PR #203, opened 2016-07-18
Support Pluggable Configuration Generators

jolynch:
Just like we can support multiple service watchers, now we can support
multiple configuration generators which can output in whatever format we
want

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ Each value in the services hash is also a hash, and must contain the following k
 
 * [`discovery`](#discovery): how synapse will discover hosts providing this service (see [below](#discovery))
 
-The services hash *should* contain a section on how to configure the routing/discovery
+The services hash *should* contain a section on how to configure the routing
 component you wish to use for this particular service. The only choice currently
 is `haproxy`:
 
@@ -146,10 +146,10 @@ is `haproxy`:
 
 The services hash may contain the following keys:
 
-* `default_servers` (default: \[\]): the list of default servers providing this service; synapse uses these if no others can be discovered. See [Listing Default Servers](#defaultservers).
+* `default_servers` (default: `[]`): the list of default servers providing this service; synapse uses these if no others can be discovered. See [Listing Default Servers](#defaultservers).
 * `keep_default_servers` (default: false): whether default servers should be added to discovered services
 * `use_previous_backends` (default: true): if at any time the registry drops all backends, use previous backends we already know about.
-<a name="backend_port_override">
+<a name="backend_port_override"/>
 * `backend_port_override`: the port that discovered servers listen on; you should specify this if your discovery mechanism only discovers names or addresses (like the DNS watcher or the Ec2TagWatcher). If the discovery method discovers a port along with hostnames (like the zookeeper watcher) this option may be left out, but will be used in preference if given.
 
 <a name="discovery"/>
@@ -250,7 +250,7 @@ It takes the following options:
 * `check_interval`: How often to request the list of tasks from Marathon (default: 10 seconds)
 * `port_index`: Index of the backend port in the task's "ports" array. (default: 0)
 
-<a name=defaultservers>
+<a name="defaultservers"/>
 #### Listing Default Servers ####
 
 You may list a number of default servers providing a service.
--- a/README.md
+++ b/README.md
@@ -141,7 +141,9 @@ Each value in the services hash is also a hash, and must contain the following k
 
 The services hash may contain the following keys:
 
-* `default_servers`: the list of default servers providing this service; synapse uses these if no others can be discovered
+* `default_servers` (default: \[\]): the list of default servers providing this service; synapse uses these if no others can be discovered. See [Listing Default Servers](#defaultservers).
+* `keep_default_servers` (default: false): whether default servers should be added to discovered services
+* `use_previous_backends` (default: true): if at any time the registry drops all backends, use previous backends we already know about.
 <a name="backend_port_override">
 * `backend_port_override`: the port that discovered servers listen on; you should specify this if your discovery mechanism only discovers names or addresses (like the DNS watcher or the Ec2TagWatcher). If the discovery method discovers a port along with hostnames (like the zookeeper watcher) this option may be left out, but will be used in preference if given.
 
@@ -243,6 +245,7 @@ It takes the following options:
 * `check_interval`: How often to request the list of tasks from Marathon (default: 10 seconds)
 * `port_index`: Index of the backend port in the task's "ports" array. (default: 0)
 
+<a name=defaultservers>
 #### Listing Default Servers ####
 
 You may list a number of default servers providing a service.
--- a/README.md
+++ b/README.md
@@ -136,7 +136,12 @@ The `services` section is a hash, where the keys are the `name` of the service t
 The name is just a human-readable string; it will be used in logs and notifications.
 Each value in the services hash is also a hash, and must contain the following keys:
 
-* [`discovery`](#discovery): how synapse will discover hosts providing this service (see below)
+* [`discovery`](#discovery): how synapse will discover hosts providing this service (see [below](#discovery))
+
+The services hash *should* contain a section on how to configure the routing/discovery
+component you wish to use for this particular service. The only choice currently
+is `haproxy`:
+
 * [`haproxy`](#haproxysvc): how will the haproxy section for this service be configured
 
 The services hash may contain the following keys:
@@ -419,3 +424,8 @@ Non-HTTP backends such as MySQL or RabbitMQ will obviously continue to need thei
 
 See the Service Watcher [README](lib/synapse/service_watcher/README.md) for
 how to add new Service Watchers.
+
+### Creating a Config Generator ###
+
+See the Config Generator [README](lib/synapse/config_generator/README.md) for
+how to add new Config Generators
--- a/README.md
+++ b/README.md
@@ -134,11 +134,16 @@ The file has three main sections.
 
 The `services` section is a hash, where the keys are the `name` of the service to be configured.
 The name is just a human-readable string; it will be used in logs and notifications.
-Each value in the services hash is also a hash, and should contain the following keys:
+Each value in the services hash is also a hash, and must contain the following keys:
 
 * [`discovery`](#discovery): how synapse will discover hosts providing this service (see below)
-* `default_servers`: the list of default servers providing this service; synapse uses these if no others can be discovered
 * [`haproxy`](#haproxysvc): how will the haproxy section for this service be configured
+
+The services hash may contain the following keys:
+
+* `default_servers`: the list of default servers providing this service; synapse uses these if no others can be discovered
+<a name="backend_port_override">
+* `backend_port_override`: the port that discovered servers listen on; you should specify this if your discovery mechanism only discovers names or addresses (like the DNS watcher or the Ec2TagWatcher). If the discovery method discovers a port along with hostnames (like the zookeeper watcher) this option may be left out, but will be used in preference if given.
 
 <a name="discovery"/>
 #### Service Discovery ####
@@ -214,9 +219,9 @@ It takes the following options:
   this is case-sensitive.
 * `tag_value`: the value to match on. Case-sensitive.
 
-Additionally, you MUST supply `server_port_override` in the `haproxy`
-section of the configuration as this watcher does not know which port
-the backend service is listening on.
+Additionally, you MUST supply [`backend_port_override`](#backend_port_override)
+in the service configuration as this watcher does not know which port the
+backend service is listening on.
 
 The following options are optional, provided the well-known `AWS_`
 environment variables shown are set. If supplied, these options will
@@ -265,7 +270,7 @@ This section is its own hash, which should contain the following keys:
 * `port`: the port (on localhost) where HAProxy will listen for connections to the service. If this is omitted, only a backend stanza (and no frontend stanza) will be generated for this service; you'll need to get traffic to your service yourself via the `shared_frontend` or manual frontends in `extra_sections`
 * `bind_address`: force HAProxy to listen on this address ( default is localhost ). Setting `bind_address` on a per service basis overrides the global `bind_address` in the top level `haproxy`. Having HAProxy listen for connections on different addresses ( example: service1 listen on 127.0.0.2:443 and service2 listen on 127.0.0.3:443) allows /etc/hosts entries to point to services.
 * `bind_options`: optional: default value is an empty string, specify additional bind parameters, such as ssl accept-proxy, crt, ciphers etc.
-* `server_port_override`: the port that discovered servers listen on; you should specify this if your discovery mechanism only discovers names or addresses (like the DNS watcher). If the discovery method discovers a port along with hostnames (like the zookeeper watcher) this option may be left out, but will be used in preference if given.
+* `server_port_override`: **DEPRECATED**. Renamed [`backend_port_override`](#backend_port_override) and moved to the top level hash. This will be removed in future versions.
 * `server_options`: the haproxy options for each `server` line of the service in HAProxy config; it may be left out.
 * `frontend`: additional lines passed to the HAProxy config in the `frontend` stanza of this service
 * `backend`: additional lines passed to the HAProxy config in the `backend` stanza of this service
--- a/README.md
+++ b/README.md
@@ -275,6 +275,9 @@ by unsetting `use_previous_backends`.
 
 This section is its own hash, which should contain the following keys:
 
+* `disabled`: A boolean value indicating if haproxy configuration management
+for just this service instance ought be disabled. For example, if you want
+file output for a particular service but no HAProxy config. (default is ``False``)
 * `port`: the port (on localhost) where HAProxy will listen for connections to the service. If this is omitted, only a backend stanza (and no frontend stanza) will be generated for this service; you'll need to get traffic to your service yourself via the `shared_frontend` or manual frontends in `extra_sections`
 * `bind_address`: force HAProxy to listen on this address ( default is localhost ). Setting `bind_address` on a per service basis overrides the global `bind_address` in the top level `haproxy`. Having HAProxy listen for connections on different addresses ( example: service1 listen on 127.0.0.2:443 and service2 listen on 127.0.0.3:443) allows /etc/hosts entries to point to services.
 * `bind_options`: optional: default value is an empty string, specify additional bind parameters, such as ssl accept-proxy, crt, ciphers etc.
--- a/lib/synapse.rb
+++ b/lib/synapse.rb
@@ -77,7 +77,7 @@ module Synapse
     end
 
     def available_generators
-      res = Hash[@config_generators.collect{|cg| [cg.name, cg]}]
+      Hash[@config_generators.collect{|cg| [cg.name, cg]}]
     end
 
     private
--- a/lib/synapse.rb
+++ b/lib/synapse.rb
@@ -19,7 +19,7 @@ module Synapse
 
       # create the service watchers for all our services
       raise "specify a list of services to connect in the config" unless opts.has_key?('services')
-      @service_watchers = create_service_watchers(opts['services'], @config_generators.keys)
+      @service_watchers = create_service_watchers(opts['services'])
 
       # configuration is initially enabled to configure on first loop
       @config_updated = true
--- a/lib/synapse.rb
+++ b/lib/synapse.rb
@@ -12,24 +12,14 @@ module Synapse
 
     include Logging
 
-    def create_config_generators(opts)
-      config_generators = []
-      opts.each do |type, generator_opts|
-        # Skip the "services" top level key
-        next if type == 'services'
-        config_generators << ConfigGenerator.create(type, generator_opts)
-      end
-      config_generators
-    end
-
     def initialize(opts={})
-      # create the service watchers for all our services
-      raise "specify a list of services to connect in the config" unless opts.has_key?('services')
-      @service_watchers = create_service_watchers(opts['services'])
-
       # create objects that need to be notified of service changes
       @config_generators = create_config_generators(opts)
       raise "no config generators supplied" if @config_generators.empty?
+
+      # create the service watchers for all our services
+      raise "specify a list of services to connect in the config" unless opts.has_key?('services')
+      @service_watchers = create_service_watchers(opts['services'], @config_generators.keys)
 
       # configuration is initially enabled to configure on first loop
       @config_updated = true
@@ -86,6 +76,10 @@ module Synapse
       @config_updated = true
     end
 
+    def available_generators
+      Hash[@config_generators.collect{|cg| [cg.name, cg]}]
+    end
+
     private
     def create_service_watchers(services={})
       service_watchers =[]
@@ -96,5 +90,16 @@ module Synapse
       return service_watchers
     end
 
+    private
+    def create_config_generators(opts={})
+      config_generators = []
+      opts.each do |type, generator_opts|
+        # Skip the "services" top level key
+        next if type == 'services'
+        config_generators << ConfigGenerator.create(type, generator_opts)
+      end
+
+      return config_generators
+    end
   end
 end
--- a/lib/synapse.rb
+++ b/lib/synapse.rb
@@ -77,7 +77,7 @@ module Synapse
     end
 
     def available_generators
-      Hash[@config_generators.collect{|cg| [cg.name, cg]}]
+      res = Hash[@config_generators.collect{|cg| [cg.name, cg]}]
     end
 
     private
--- a/lib/synapse.rb
+++ b/lib/synapse.rb
@@ -82,7 +82,7 @@ module Synapse
 
     private
     def create_service_watchers(services={})
-      service_watchers =[]
+      service_watchers = []
       services.each do |service_name, service_config|
         service_watchers << ServiceWatcher.create(service_name, service_config, self)
       end
--- a/lib/synapse.rb
+++ b/lib/synapse.rb
@@ -1,11 +1,10 @@
 require 'logger'
 require 'json'
 
-require "synapse/version"
-require "synapse/log"
-require "synapse/haproxy"
-require "synapse/file_output"
-require "synapse/service_watcher"
+require 'synapse/version'
+require 'synapse/log'
+require 'synapse/config_generator'
+require 'synapse/service_watcher'
 
 
 module Synapse
@@ -13,22 +12,24 @@ module Synapse
 
     include Logging
 
+    def create_config_generators(opts)
+      config_generators = []
+      opts.each do |type, generator_opts|
+        # Skip the "services" top level key
+        next if type == 'services'
+        config_generators << ConfigGenerator.create(type, generator_opts)
+      end
+      config_generators
+    end
+
     def initialize(opts={})
       # create the service watchers for all our services
       raise "specify a list of services to connect in the config" unless opts.has_key?('services')
       @service_watchers = create_service_watchers(opts['services'])
 
       # create objects that need to be notified of service changes
-      @config_generators = []
-      # create the haproxy config generator, this is mandatory
-      raise "haproxy config section is missing" unless opts.has_key?('haproxy')
-      @config_generators << Haproxy.new(opts['haproxy'])
-
-      # possibly create a file manifestation for services that do not
-      # want to communicate via haproxy, e.g. cassandra
-      if opts.has_key?('file_output')
-        @config_generators << FileOutput.new(opts['file_output'])
-      end
+      @config_generators = create_config_generators(opts)
+      raise "no config generators supplied" if @config_generators.empty?
 
       # configuration is initially enabled to configure on first loop
       @config_updated = true
--- a/lib/synapse/config_generator.rb
+++ b/lib/synapse/config_generator.rb
@@ -1,0 +1,20 @@
+require 'synapse/log'
+require 'synapse/config_generator/base'
+
+module Synapse
+  class ConfigGenerator
+    # the method which actually dispatches generator creation requests
+    def self.create(type, opts)
+      generator = begin
+        method = type.downcase
+        require "synapse/config_generator/#{method}"
+        # haproxy => Haproxy, file_output => FileOutput, etc ...
+        method_class  = method.split('_').map{|x| x.capitalize}.join
+        self.const_get("#{method_class}")
+      rescue Exception => e
+        raise ArgumentError, "Specified a config generator of #{method}, which could not be found: #{e}"
+      end
+      return generator.new(opts)
+    end
+  end
+end
--- a/lib/synapse/config_generator.rb
+++ b/lib/synapse/config_generator.rb
@@ -3,16 +3,16 @@ require 'synapse/config_generator/base'
 
 module Synapse
   class ConfigGenerator
-    # the method which actually dispatches generator creation requests
+    # the type which actually dispatches generator creation requests
     def self.create(type, opts)
       generator = begin
-        method = type.downcase
-        require "synapse/config_generator/#{method}"
+        type = type.downcase
+        require "synapse/config_generator/#{type}"
         # haproxy => Haproxy, file_output => FileOutput, etc ...
-        method_class  = method.split('_').map{|x| x.capitalize}.join
-        self.const_get("#{method_class}")
+        type_class  = type.split('_').map{|x| x.capitalize}.join
+        self.const_get("#{type_class}")
       rescue Exception => e
-        raise ArgumentError, "Specified a config generator of #{method}, which could not be found: #{e}"
+        raise ArgumentError, "Specified a config generator of #{type}, which could not be found: #{e}"
       end
       return generator.new(opts)
     end
--- a/lib/synapse/config_generator/README.md
+++ b/lib/synapse/config_generator/README.md
@@ -1,11 +1,11 @@
 ## ConfigGenerator Classes
 
-Generators are the piece of Synapse that react to changes in Generators and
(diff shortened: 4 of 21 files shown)


jolynch:
@igor47 This PR is the start of a refactor that I hope will make it so that Synapse can take registrations from any registry and make them appear in whichever load balancing solution customers want.

**Why?**
HAProxy has some drawbacks for simple applications that don't need super powerful load balancing features and the community is definitely moving to polyglot load balancing solutions. In particular, nginx is starting to look much more viable as the open source version supports more and more load balancing features (and hitless reloads out of the box). Also it's always been super gross IMO that service watchers are coupled through to haproxy, so I'm trying to clean that up a bit.

**What's Done?**
Just like watchers folks can dynamically add ``ConfigGenerators`` which respond to ``config_updates`` calls from synapse. I have cleaned up the interface a bit.

**What's TODO?**
I have to figure out how to decouple [haproxy](https://github.com/airbnb/synapse/blob/64ddbbd6140d337d65cf0f31fff9ea004aca779e/lib/synapse/service_watcher/base.rb#L29-L35) from the ServiceWatchers. I'm thinking of exposing the generators through a function on the ``@synapse`` object that ServiceWatchers get at construction, and maybe do the validation there. The thing I'm primarily struggling with is that things like the Docker watcher reach into the haproxy implementation to figure out which port they need. I'm working on decoupling that as well.

jolynch:
@igor47 so this is one strategy that I think will work. I'm working on cleaning up the specs and code, but the current diff shows my current idea (using the synapse object to communicate between the two abstractions, but only in a limited fashion).

jolynch:
@igor47, @jnb  Alright, I think that this is in a fairly reasonable state for review :-D

Manual tests and specs give me reasonable confidence this works, but I can always add more specific testing if you like.

README.md
```diff
@@ -134,12 +134,24 @@ The file has three main sections.
 
 The `services` section is a hash, where the keys are the `name` of the service to be configured.
 The name is just a human-readable string; it will be used in logs and notifications.
-Each value in the services hash is also a hash, and should contain the following keys:
+Each value in the services hash is also a hash, and must contain the following keys:
+
+* [`discovery`](#discovery): how synapse will discover hosts providing this service (see [below](#discovery))
+
+The services hash *should* contain a section on how to configure the routing/discovery
```
igor47: s|routing/discovery|routing ?

README.md
```diff
@@ -134,12 +134,24 @@ The file has three main sections.
...
+component you wish to use for this particular service. The only choice currently
+is `haproxy`:
 
-* [`discovery`](#discovery): how synapse will discover hosts providing this service (see below)
-* `default_servers`: the list of default servers providing this service; synapse uses these if no others can be discovered
 * [`haproxy`](#haproxysvc): how will the haproxy section for this service be configured
 
+The services hash may contain the following keys:
+
+* `default_servers` (default: \[\]): the list of default servers providing this service; synapse uses these if no others can be discovered. See [Listing Default Servers](#defaultservers).
```
igor47: maybe "default: \`[]\`" ?

README.md
```diff
@@ -238,6 +250,7 @@ It takes the following options:
 * `check_interval`: How often to request the list of tasks from Marathon (default: 10 seconds)
 * `port_index`: Index of the backend port in the task's "ports" array. (default: 0)
 
+<a name=defaultservers>
```
igor47: should these `<a` have a `/>` to close the tags?

lib/synapse.rb
```diff
@@ -85,15 +76,30 @@ def reconfigure!
       @config_updated = true
     end
 
+    def available_generators
+      res = Hash[@config_generators.collect{|cg| [cg.name, cg]}]
```
igor47: `res` is unused

lib/synapse/config_generator/haproxy.rb
```diff
@@ -792,8 +795,6 @@ class Haproxy
     STATE_FILE_UPDATE_INTERVAL = 60.freeze # iterations; not a unit of time
 
     def initialize(opts)
-      super()
-
       %w{global defaults reload_command}.each do |req|
         raise ArgumentError, "haproxy requires a #{req} section" if !opts.has_key?(req)
       end
```
jolynch: Not related to this PR, but we should make sure we fix the ordering of req_pairs and setting default values for do_writes/do_socket/do_reloads. I'll fix that in the next iteration.

lib/synapse/config_generator/README.md
```diff
@@ -0,0 +1,64 @@
+## ConfigGenerator Classes
+
+Generators are the piece of Synapse that react to changes in service
+registrations and actually reflect those changes in local state.
+Generators should conform to the interface specified by `BaseGenerator` and
+when your generator has received an update synapse via `update_config` it
```
igor47: an update *from* synapse?

igor47:
ugh sorry i've been so slow on reviewing this. this PR is very relevant to my interests, but it's been kind of a crazy week. i'm about half-done and i'll finish going through it tomorrow

jolynch:
No problem, I've been mostly hacking on this for fun so there isn't too much time pressure (the nerve change actually is hurting us right now heh). Mostly doing this because I want to allow services to be "discovery only" like cassandra which we will never talk through HAProxy for but we are still managing it in HAProxy because Synapse requires it to be managed in HAProxy heh. 

Separately some folks at work are huge nginx fanboys ("it reloads without dropping traffic", "it can do retries of requests", "it can scale to more cores", etc ...) and I wanted to see what the reality was. Well after writing [this proof of concept](https://gist.github.com/jolynch/96045a86a2ae5af62c956830cca9b77c) it does seem like it _could_ work, but we'd have to reload for every up/down/flapped healthcheck (since it doesn't have the equivalent of a stats socket unless you pay for nginx+). I think that I might be comfortable putting marathon based HTTP services through it though because those already require restarts on every state change (marathon is a jerk about keeping ports the same) and they typically want non spec compliant things like retry-ing HTTP requests.

README.md
jolynch: Sure, I view the file_output as more discovery than routing (unlike haproxy which is both), but sure.

README.md
jolynch: done

README.md
jolynch: indeed, on it

lib/synapse.rb
jolynch: oops, leftover from some debugging

lib/synapse/config_generator/haproxy.rb
jolynch: Alright, I cleaned this up and added some more (regression?) tests

lib/synapse/config_generator/README.md
jolynch: done

README.md
```diff
@@ -134,12 +134,24 @@ The file has three main sections.
...
 * [`haproxy`](#haproxysvc): how will the haproxy section for this service be configured
 
+The services hash may contain the following keys:
+
+* `default_servers` (default: `[]`): the list of default servers providing this service; synapse uses these if no others can be discovered. See [Listing Default Servers](#defaultservers).
+* `keep_default_servers` (default: false): whether default servers should be added to discovered services
+* `use_previous_backends` (default: true): if at any time the registry drops all backends, use previous backends we already know about.
+<a name="backend_port_override"/>
+* `backend_port_override`: the port that discovered servers listen on; you should specify this if your discovery mechanism only discovers names or addresses (like the DNS watcher or the Ec2TagWatcher). If the discovery method discovers a port along with hostnames (like the zookeeper watcher) this option may be left out, but will be used in preference if given.
+
```
jolynch: I've noticed that this isn't actually a full list. Let's document all the options that we can set (filters, leader_election, etc ...)

jolynch:
@igor47 now that envoy seems to be out and about (and it looks pretty sweet), what are your thoughts on this PR?

jolynch:
@igor47 @jnb gently poke

I think this branch is good for merge and I'm about 80% of the way through a nginx adding branch that will give us an e2e test of this functionality (synapse managing multiple proxies simultaneously).

@vulpine @bobtfish feel free to review if you want nginx fronted haproxy goodness :-D

lib/synapse/service_watcher/base.rb
```diff
@@ -33,16 +33,10 @@ def initialize(opts={}, synapse)
...
-      @haproxy['server_port_override'] ||= nil
-      %w{backend frontend listen}.each do |sec|
-        @haproxy[sec] ||= []
-      end
-
-      unless @haproxy.include?('port')
-        log.warn "synapse: service #{name}: haproxy config does not include a port; only backend sections for the service will be created; you must move traffic there manually using configuration in `extra_sections`"
+      @generator_config = {}
+      @synapse.available_generators.each do |generator_name, generator|
+        @generator_config[generator_name] = opts[generator_name] || {}
```
igor47: i would prefer a more functional declaration:

```ruby
def generator_config
  @generator_config ||= Hash[@synapse.available_generators.map do |gen_name, gen|
    gen_opts = opts[gen_name] || {}
    normalized_opts = gen.normalize_config_generator_opts!(gen_opts)
    [gen_name, normalized_opts]
  end]
end
```

this might require us to persist the opts in the initializer:
```ruby
attr_reader :opts
def initialize(opts = {}, synapse)
  @opts = opts
end
```

lib/synapse/config_generator/base.rb
```diff
@@ -0,0 +1,37 @@
...
+    # or restarts in a rate limited fashion
+    def tick
+    end
+
+    # Service watchers have a subsection of their ``services`` entry that is
+    # dedicated to the watcher specific configuration for how to configure
+    # the config generator. This method will be called with each of these
+    # watcher hashes, and should normalize them to what the config generator
+    # needs, such as adding defaults. Mutate the passed hash
+    def normalize_config_generator_opts!(service_watcher_name, service_watcher_opts)
```
igor47: i was confused by there being two sources of options for a config generator. the first is the options passed to the `initialize` method, which contain overall options. the second is the watcher-specific options for this config generator.

maybe we should call these options `watcher_opts` rather than `config_generator_opts` to clarify the distinction?

also, should we really be encouraging side-effects? this might lead to trouble. given how we actually use this code below, returning a copy of the passed-in hash would work just as well. we should encourage users to do something like:
```ruby
def normalize_watcher_opts(watcher_name, watcher_opts)
  defaults = {}
  normalized_opts = watcher_opts.merge(defaults)
end
```

finally, i was also confused by the fact that we have to reach into the watchers to get the config generator options each time we generate a config for that watcher. i guess i'm not sure what to do about that problem, but i'm kind of annoyed about repeated calls to, e.g., `watcher.generator_opts['haproxy']`

lib/synapse/config_generator/README.md
```diff
@@ -0,0 +1,64 @@
...
+```ruby
+require "synapse/config_generator/base"
+
+class Synapse::ServiceGenerator
+  class MyGenerator < BaseGenerator
+    # The generator name is used to find service specific
+    # configuration in the service watchers. When supplying
+    # per service config, use this as the key
+    NAME = 'my_generator'.freeze
+
```
igor47: we should document `initialize`:

```ruby
# The generator will be passed-in any options specified in it's dedicated section. Those are
# given as an `opts` hash to the initializer. You may (and probably should) omit this method,
# or you can declare your own -- remember to invoke the parent initializer!
def initialize(opts = {})
  super(opts)
end
```

lib/synapse/config_generator/base.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    NAME = 'base'.freeze
+
+    # Exposes NAME as 'name' so we can remain consistent with how we refer to
+    # service_watchers' by generator.name access (instead of
+    # generator.class::NAME) even though the names of generators don't change
+    def name
+      self.class::NAME
+    end
+
```
igor47: should we add a default initializer?
```ruby
attr_reader :opts
def initialize(opts = {})
  @opts = opts
end
```

lib/synapse/config_generator/haproxy.rb
```diff
@@ -961,17 +973,28 @@ def validate_haproxy_stanza(stanza, stanza_type, service_name)
 
     # generates an individual stanza for a particular watcher
     def generate_frontend_stanza(watcher, config)
-      unless watcher.haproxy.has_key?("port")
+      unless watcher.generator_config['haproxy'].has_key?("port")
```
igor47: can we declare a `watcher_opts` var so we don't have to keep calling `watcher.generator_config['haproxy']`? also, should we be hard-coding `'haproxy'`? i think we should use `name`:

```ruby
watcher_opts = watcher.generator_config[name]
```

lib/synapse/config_generator/haproxy.rb
```diff
@@ -961,17 +973,28 @@ def validate_haproxy_stanza(stanza, stanza_type, service_name)
...
         log.debug "synapse: not generating frontend stanza for watcher #{watcher.name} because it has no port defined"
         return []
+      else
+        port = watcher.generator_config['haproxy']['port']
```
igor47: we can probably do:
```ruby
port = watcher_opts['port']
unless port
end
```

lib/synapse/config_generator/haproxy.rb
```diff
@@ -1005,7 +1028,8 @@ def generate_backend_stanza(watcher, config)
         log.debug "synapse: no backends found for watcher #{watcher.name}"
       end
 
-      keys = case watcher.haproxy['backend_order']
+      watcher_haproxy = watcher.generator_config['haproxy']
```
igor47: whenever we do this, lets keep the name consistent. i propose `watcher_opts` as a variable that generally refers to the watcher-specific options for the current generator

lib/synapse/config_generator/haproxy.rb
```diff
@@ -1151,9 +1176,9 @@ def restart
       @next_restart += rand(@restart_jitter * @restart_interval + 1)
 
       # do the actual restart
-      res = `#{opts['reload_command']}`.chomp
+      res = `#{@opts['reload_command']}`.chomp
```
igor47: i feel like we should be moving towards accessor methods and away from explicit instance vars, not the other way. this makes testing easier, since you can `allow(generator).to receive(:opts).and_return({})` rather than resorting to meta-programming in the specs (`generator.instance_variable_set(:ops, {})`)

lib/synapse/service_watcher/base.rb
```diff
@@ -7,15 +7,15 @@ class BaseWatcher
 
     LEADER_WARN_INTERVAL = 30
 
-    attr_reader :name, :haproxy
+    attr_reader :name, :generator_config
```
igor47: should `generator_config` be  plural?

lib/synapse/service_watcher/base.rb
```diff
@@ -55,13 +49,28 @@ def initialize(opts={}, synapse)
...
       # set a flag used to tell the watchers to exit
       # this is not used in every watcher
       @should_exit = false
 
       validate_discovery_opts
     end
 
+    def haproxy
+      log.warn "synapse: service watcher #{@name} accessing watcher.haproxy. This is DEPRECATED and will be removed in future iterations, use watcher.generator_config['haproxy'] instead."
+      @generator_config['haproxy']
```
igor47: image

lib/synapse/service_watcher/base.rb
```diff
@@ -55,13 +49,28 @@ def initialize(opts={}, synapse)
       # use the previous backends that we already know about.
       @use_previous_backends = opts.fetch('use_previous_backends', true)
 
+      # If backends discovered with this watcher lack ports or discovered
+      # ports should be ignored, this option indicates that
+      @backend_port_override = opts['backend_port_override'] || nil
+
+      # For backwards compatability we support server_port_override
+      # This will be removed in future versions
```
igor47: :+1:

lib/synapse.rb
```diff
@@ -85,15 +76,30 @@ def reconfigure!
...
     end
 
+    def available_generators
+      Hash[@config_generators.collect{|cg| [cg.name, cg]}]
+    end
+
     private
     def create_service_watchers(services={})
-      service_watchers =[]
+      service_watchers = []
```
igor47: 💯

lib/synapse/service_watcher/base.rb
```diff
@@ -55,13 +49,28 @@ def initialize(opts={}, synapse)
...
+
       # set a flag used to tell the watchers to exit
       # this is not used in every watcher
       @should_exit = false
 
       validate_discovery_opts
     end
 
+    def haproxy
+      log.warn "synapse: service watcher #{@name} accessing watcher.haproxy. This is DEPRECATED and will be removed in future iterations, use watcher.generator_config['haproxy'] instead."
```
igor47: we're going to end up logging this warning quite a lot. maybe we should only do it once?

lib/synapse/service_watcher/ec2tag.rb
```diff
@@ -36,13 +36,13 @@ def validate_discovery_opts
...
-      unless @haproxy['server_port_override']
+      unless @backend_port_override
         raise ArgumentError,
-          "Missing server_port_override for service #{@name} - which port are backends listening on?"
+          "Missing backend_port_override for service #{@name} - which port are backends listening on?"
       end
 
-      unless @haproxy['server_port_override'].to_s.match(/^\d+$/)
-        raise ArgumentError, "Invalid server_port_override value"
+      unless @backend_port_override.to_s.match(/^\d+$/)
```
igor47: should this validation, too, live in `BaseWatcher` ?

lib/synapse/config_generator.rb
```diff
@@ -0,0 +1,20 @@
+require 'synapse/log'
+require 'synapse/config_generator/base'
+
+module Synapse
+  class ConfigGenerator
+    # the type which actually dispatches generator creation requests
+    def self.create(type, opts)
+      generator = begin
+        type = type.downcase
+        require "synapse/config_generator/#{type}"
```
jolynch: I was under the impression this is already possible if you install via gems? All you have to do is put your class in the right path and then synapse dynamically requires it, or that's what I thought the design from Nerve allowed you to do.

I can whip up a quick example gem to see if this is flawed thinking, but I thought that was why we switched to this way of constructing classes in Nerve (and Synapse).

lib/synapse/config_generator/README.md
jolynch: Done.

lib/synapse/config_generator/README.md
```diff
@@ -0,0 +1,64 @@
...
+    def update_config(watchers)
+      # synapse will call this method whenever watcher state changes with the
+      # watcher state. You should reflect that state in the local config state
+    end
+
+    def tick
+      # Called every loop of the main Synapse loop regardless of watcher #
```
jolynch: fixed

lib/synapse/config_generator/README.md
```diff
@@ -0,0 +1,64 @@
...
+      # changes (roughly ~1s), you can use this to rate limit how often your
```
jolynch: Fixed

lib/synapse/config_generator/README.md
```diff
@@ -0,0 +1,64 @@
...
+      # config generator actually reconfigures external services (e.g. HAProxy
+      # may need to rate limit reloads as those can be disruptive to in
+      # flight connections
+    end
+
+    def normalize_config_generator_opts!(service_watcher_name, service_watcher_opts)
+      # Every service section can contain configuration that changes how the
+      # config generator reacts for that particular service. Typically this
```
jolynch: Reworded, let me know if it's clearer :-)

lib/synapse/config_generator/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+#### Class Name
+These type strings are then transformed into class names via the following
+function:
+
+```
+type_class  = type.split('_').map{|x| x.capitalize}
+```
+
+This has the effect of taking the method, splitting on '_', capitalizing each
```
jolynch: Fixed

lib/synapse/config_generator/base.rb
jolynch: Sure, done.

lib/synapse/config_generator/base.rb
```diff
@@ -0,0 +1,37 @@
...
+    # The synapse main loop will call this any time watchers change, the
+    # config_generator is responsible for diffing the passed watcher state
+    # against the output configuration
+    def update_config(watchers)
```
jolynch: I don't think it's needed. Ok to punt?

lib/synapse/config_generator/base.rb
jolynch: > i was confused by there being two sources of options for a config generator. the first is the options passed to the initialize method, which contain overall options. the second is the watcher-specific options for this config generator.

Yup, the fact that there is coupling between the two presented a significant challenge during this refactor.

>maybe we should call these options watcher_opts rather than config_generator_opts to clarify the distinction?
Hm, I will work on making this clearer

> also, should we really be encouraging side-effects? this might lead to trouble. given how we actually use this code below, returning a copy of the passed-in hash would work just as well. we should encourage users to do something like:

Yea, I think I can figure that out

>finally, i was also confused by the fact that we have to reach into the watchers to get the config generator options each time we generate a config for that watcher. i guess i'm not sure what to do about that problem, but i'm kind of annoyed about repeated calls to, e.g., watcher.generator_opts['haproxy']

Yes, I agree, but it's not that bad and we can refactor this in the future imo.

lib/synapse/config_generator/haproxy.rb
```diff
@@ -792,29 +795,27 @@ class Haproxy
     STATE_FILE_UPDATE_INTERVAL = 60.freeze # iterations; not a unit of time
 
     def initialize(opts)
-      super()
-
-      %w{global defaults reload_command}.each do |req|
+      %w{global defaults}.each do |req|
```
jolynch: I'd like to punt this to another iteration.

lib/synapse/config_generator/haproxy.rb
jolynch: Sure, I already did this for nginx, but was punting on this one. I can make the changes now though.

lib/synapse/config_generator.rb
igor47: oh i see. yeah, i guess if you `gem install synapse` then it would work with any other systemwide-installed gems. if you install a synapse bundle and `bundle exec synapse`, you have to put the other gem system-wide or bundler won't find it

lib/synapse/config_generator/base.rb
igor47: sure

lib/synapse/config_generator/haproxy.rb
jolynch: So ... I have plans over in #220 to allow a null port to be semantically different from no port. Basically for unix socket listeners I need to not supply a port, but we still want the frontend and to use bind_address.

lib/synapse/config_generator/haproxy.rb
jolynch: I ended up going with watcher_config, but feel free to rename to opts if you want (opts and config are hella confusing no matter how we slice it :-/)

lib/synapse/config_generator/haproxy.rb
jolynch: Agreed, changed.

lib/synapse.rb
```diff
@@ -85,15 +76,30 @@ def reconfigure!
       @config_updated = true
     end
 
+    def available_generators
+      Hash[@config_generators.collect{|cg| [cg.name, cg]}]
```
jolynch: This method is only called during initialization of watchers. I don't really want to add state here unless I have to.

lib/synapse/service_watcher/base.rb
jolynch: I don't think so but ... naming is hard?

lib/synapse/service_watcher/base.rb
jolynch: Ok, I've made it more functional but I didn't want to make it a separate function because I don't really want to persist opts. Punt for now?

lib/synapse/service_watcher/base.rb
jolynch: This was really just to help me when porting the service watchers. I think this would only impact folks if they had custom service watchers (all of the ones in the repo have been ported).

lib/synapse/service_watcher/base.rb
jolynch: heh, ack

lib/synapse/service_watcher/ec2tag.rb
jolynch: Sure, moved

jolynch:
@igor47 I took another pass, let me know if there are any remaining things you want me to fix in this iteration :-)

spec/lib/synapse/haproxy_spec.rb
```diff
@@ -1,16 +1,19 @@
 require 'spec_helper'
+require 'synapse/config_generator/haproxy'
```
igor47: this spec file should probably live in `spec/lib/config_generator/haproxy_spec.rb`

lib/synapse/config_generator/base.rb
```diff
@@ -0,0 +1,44 @@
...
+    def tick
+    end
+
+    # Service watchers have a subsection of their ``services`` entry that is
+    # dedicated to the watcher specific configuration for how to configure
+    # the config generator. This method will be called with each of these
+    # watcher hashes, and should normalize them to what the config generator
+    # needs, such as adding defaults. Return the properly populated default hash
+    def normalize_watcher_provided_opts(service_watcher_name, service_watcher_opts)
+      service_watcher_opts
```
igor47: should we `.dup` this?

lib/synapse/service_watcher/base.rb
```diff
@@ -33,17 +33,15 @@ def initialize(opts={}, synapse)
...
-      @haproxy['server_options'] ||= ""
-      @haproxy['server_port_override'] ||= nil
-      %w{backend frontend listen}.each do |sec|
-        @haproxy[sec] ||= []
-      end
-
-      unless @haproxy.include?('port')
-        log.warn "synapse: service #{name}: haproxy config does not include a port; only backend sections for the service will be created; you must move traffic there manually using configuration in `extra_sections`"
-      end
+      @generator_config = Hash[
```
igor47: so, this would become `@config_for_generator` if you accept my earlier suggestion

lib/synapse/config_generator/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+    def tick
+      # Called every loop of the main Synapse loop regardless of watcher
+      # changes (roughly ~1s). You can use this to rate limit how often your
+      # config generator actually reconfigures external services (e.g. HAProxy
+      # may need to rate limit reloads as those can be disruptive to in
+      # flight connections
+    end
+
+    def normalize_watcher_provided_opts(service_watcher_name, service_watcher_opts)
```
igor47: we should either call this `opts` or `config`. seems like we settled on `config`

lib/synapse.rb
igor47: eh, i feel like it's pretty standard practice to memoize in ruby if you need a specific view of otherwise-existing state. but UP2U!

spec/lib/synapse/haproxy_spec.rb
```diff
@@ -28,37 +33,165 @@ class MockWatcher; end;
...
+    allow(mockWatcher).to receive(:name).and_return('disabled_watcher')
+    backends = [{ 'host' => 'somehost', 'port' => 5555}]
+    allow(mockWatcher).to receive(:backends).and_return(backends)
+    allow(mockWatcher).to receive(:generator_config).and_return({
+      'haproxy' => {'port' => 2200, 'disabled' => true}
+    })
     mockWatcher
   end
 
+  describe 'validates arguments' do
```
igor47: should this be called `describe '#initialize'` ?

spec/lib/synapse/haproxy_spec.rb
```diff
@@ -28,37 +33,165 @@ class MockWatcher; end;
...
+    end
+
+    it 'does not generate config' do
+      allow(subject).to receive(:parse_watcher_config).and_return({})
+      expect(subject).to receive(:generate_frontend_stanza).exactly(:once).with(mockwatcher, nil)
+      expect(subject).to receive(:generate_backend_stanza).exactly(:once).with(mockwatcher, nil)
+      subject.update_config(watchers)
+    end
+
+    it 'does not cause a restart due to the socket' do
```
igor47: maybe more like:

```ruby
context 'when configuration via the socket succeeds' do
  before do
    # make the context true by stubbing out `talk_to_socket`
  end

  it 'does not force a restart' do
    subject.update_config(watchers)
    expect(...).to be_falsey
  end
end
```

lib/synapse/config_generator/haproxy.rb
```diff
@@ -877,14 +890,19 @@ def generate_config(watchers)
       shared_frontend_lines = generate_shared_frontend
 
       watchers.each do |watcher|
+        watcher_config = watcher.generator_config[name]
```
jolynch: I like that better too. Done.

lib/synapse/config_generator/haproxy.rb
```diff
@@ -961,17 +980,29 @@ def validate_haproxy_stanza(stanza, stanza_type, service_name)
...
+      unless watcher_config.has_key?("port")
         log.debug "synapse: not generating frontend stanza for watcher #{watcher.name} because it has no port defined"
         return []
+      else
+        port = watcher_config['port']
       end
 
-      bo = watcher.haproxy['bind_options']
+      bind_options = watcher_config['bind_options']
+      bind_options = bind_options.nil? ? '' : ' ' + bind_options
```
jolynch: Sure, done. I can also fixup the address:port logic in the nginx review as well to do a similar thing.

lib/synapse/config_generator/haproxy.rb
```diff
@@ -961,17 +980,29 @@ def validate_haproxy_stanza(stanza, stanza_type, service_name)
...
+
+      bind_address = (
+        watcher_config['bind_address'] ||
+        opts['bind_address'] ||
+        'localhost'
```
jolynch: Sure, pre-existing but sure.

lib/synapse/service_watcher/base.rb
jolynch: Yup

lib/synapse/service_watcher/base.rb
```diff
@@ -33,17 +33,15 @@ def initialize(opts={}, synapse)
...
+        @synapse.available_generators.collect{ |generator_name, generator|
```
jolynch: Done

lib/synapse/service_watcher/base.rb
```diff
@@ -135,6 +154,11 @@ def backends_filtered_by_labels
     def set_backends(new_backends)
       # Aggregate and deduplicate all potential backend service instances.
       new_backends = (new_backends + @default_servers) if @keep_default_servers
+      # Substitute backend_port_override for the provided port
+      new_backends = new_backends.collect {|b|
```
jolynch: Done

spec/lib/synapse/file_output_spec.rb
```diff
@@ -1,8 +1,9 @@
 require 'spec_helper'
 require 'fileutils'
+require 'synapse/config_generator/file_output'
```
jolynch: I agree, but right now in synapse we don't group specs yet. I think punt and followup.

spec/lib/synapse/service_watcher_base_spec.rb
```diff
@@ -5,7 +5,14 @@ class Synapse::ServiceWatcher::BaseWatcher
 end
 
 describe Synapse::ServiceWatcher::BaseWatcher do
-  let(:mocksynapse) { double() }
+  let(:mocksynapse) do
```
jolynch: Sure, but I think we can do that in a follow up.

spec/lib/synapse/haproxy_spec.rb
jolynch: Let's clean up the specs in a separate change.

spec/lib/synapse/haproxy_spec.rb
jolynch: Don't have much of an opinion, renamed.

spec/lib/synapse/haproxy_spec.rb
jolynch: Sure, I made it a context, except that we can't stub out talk_to_socket the same way for the two specs (one where the server was previously in the backends, and one without).

Also I don't like testing for falsey when that's not something that is supposed to be ok, namely nil isn't a valid value at all for restart_required.

lib/synapse/config_generator/base.rb
jolynch: Sure

lib/synapse/config_generator/README.md
jolynch: Ack

jolynch:
@igor47 I agree more refactoring is possible (especially in the specs overall), but I'd like to get this into master and start getting production experience with it asap since it's just a major refactor (and likely to have subtle bugs).

Are you ok to merge and then I can follow up with any bugs I find (and specs to match)?

igor47:
ok, merging. i'll try to get this into some prod roles at airbnb this week, and i'll also take a look at the specs

jolynch:
Sweet thanks, I'll start rolling this out here ASAP and making sure I didn't break anything :-/

Issue #205, opened 2016-08-03
ZookeeperWatcher does not handle NoNode failures


jolynch:
There is a race condition in how the ZookeeperWatcher [discovers backends](https://github.com/airbnb/synapse/blob/1edd876513b850cda93fb20d68226c69682973ae/lib/synapse/service_watcher/zookeeper.rb#L129-L130) where we can crash the whole Synapse process if we try to read a node that doesn't exist after the list. For example:

```
E, [2016-08-02T17:02:32.909587 #113325] ERROR -- Synapse::Synapse: synapse: encountered unexpected exception #<ZK::Exceptions::NoNode: inputs: {:path=>"/nerve/..."}> in main thread
```

We've seen this cause stability issues in production, let's catch that NoNode exception and continue.

igor47:
i guess what would happen is synapse would bail and restart. this would force a restart of haproxy which might cause some instability. we've certainly spent a lot of time figuring out how to limit the frequency of haproxy restarts.

i've thought about not restarting haproxy when we boot. but this is not possible. we control a lot of haproxy settings which we don't and can't check for via the stats socket. restarting after writing the config is the only way to be sure that we've probably set random httpcheck options or like shared frontends or queue settings.

jolynch:
cut #206 to mitigate

Interestingly the issue for us was that a few particular Synapse instances kept crashing rather than making it through to configuration, which led to HAProxy having all old backends.

@igor47 Have you guys tried ``state_file_path`` with ``state_file_ttl`` as well as ``restart_interval`` and ``restart_jitter``? Those options were able to get our HAProxy reloads pretty much under control. I think we could limit it even on startup by applying the [jitter logic](https://github.com/airbnb/synapse/blob/master/lib/synapse/haproxy.rb#L1149-L1150) to the init of [@next_restart](https://github.com/airbnb/synapse/blob/master/lib/synapse/haproxy.rb#L829). I think that @solarkennedy mentioned he might be interested in making that change.

igor47:
we use `state_file_path`, and this helps quite a bit. the only case where we restart today is when synapse restarts and when we have to add an entirely new backend.

i was trying to say that i think we *must* restart haproxy when we restart synapse.

jolynch:
[...]

PR #206, opened 2016-08-03
Resilient zk watcher

jolynch:
This should mitigate #205

### DIFF
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -127,7 +127,12 @@ class Synapse::ServiceWatcher
 
       new_backends = []
       @zk.children(@discovery['path'], :watch => true).each do |id|
-        node = @zk.get("#{@discovery['path']}/#{id}")
+        begin
+          node = @zk.get("#{@discovery['path']}/#{id}")
+        rescue ZK::Exceptions::NoNode => e
+          log.error("synapse: consistency issue reading #{id} at #{@discovery['path']}: #{e}")
+          next
+        end
 
         begin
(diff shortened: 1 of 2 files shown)


lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -127,7 +127,12 @@ def discover
 
       new_backends = []
       @zk.children(@discovery['path'], :watch => true).each do |id|
-        node = @zk.get("#{@discovery['path']}/#{id}")
+        begin
+          node = @zk.get("#{@discovery['path']}/#{id}")
+        rescue ZK::Exceptions::NoNode => e
+          log.error("synapse: consistency issue reading #{id} at #{@discovery['path']}: #{e}")
```
igor47: i guess the idea here is that we found the node in the list, but then it disappeared before we could read it's contents. so, we will omit it from `new_backends`. should we make the log message more explanatory? people are going to be confused by what a `consistency issue` is, but it sounds pretty scary rather than routine.

igor47:
looks good, i left one nit about the wording of the log message.

lib/synapse/service_watcher/zookeeper.rb
igor47: something like `"synapse: node #{id} disappeared before it could be read"` is probably fine, along with a comment in the code explaining why this could happen.

jolynch: ack, changed.

igor47:
should we push out a new version?

jolynch:
@igor47 Ack pushed out 0.13.8

PR #207, opened 2016-08-09
[base watcher] filter on set_backends

igor47:
currently, we are filtering backends when we attempt to retrieve the list of
backends. there are a few things wrong with that:
- we perform the same filter computation repeatedly every time we rebuild the haproxy config
- we incorrectly restart haproxy even if the set of backends which don't match the filters changes
- we log the total number of backends discovered, not the number of backends matching the filter
- it's unclear how the defaults/keep_last logic plays with filters/leader election

this PR moves the filter/leader election logic to be performed as part of
`set_backends`. the call to retrieve the backends becomes just an
`attr_reader`. we don't have to make any spec changes, because we already test
all of this logic as part of testing `set_backends`

this change may be backwards-incompatible depending on configuration, because
the logic around how default/previous backends are used is different.
previously, if no backends match filters (including the default backends) or if
no backends are elected as leader (including default backends) then we return
an empty list. after this change is made, we will apply the filters/election
parameters to just the discovered backends. if these fail election or
filtering, then we will return the default backends instead of returning an
empty list.

i think this change makes things more clear overall. previously, there was no
mechanism to always return some default set of backends no matter what, and now
this mechanism is just the setting of `default_backends`

to: @jtai @jolynch @scarletmeow

### DIFF
--- a/lib/synapse/service_watcher/base.rb
+++ b/lib/synapse/service_watcher/base.rb
@@ -7,7 +7,7 @@ class Synapse::ServiceWatcher
 
     LEADER_WARN_INTERVAL = 30
 
-    attr_reader :name, :haproxy
+    attr_reader :name, :haproxy, :backends
 
     def initialize(opts={}, synapse)
       super()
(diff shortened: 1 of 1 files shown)


jtai:
I'm not sure about the backwards incompatibility concerns, but the code change looks good to me.

Issue #78, opened 2014-07-06
Be smarter about initial reconfigure


dcosson:
Synapse is designed to re-write the haproxy config & reload haproxy when it starts (set [here](https://github.com/airbnb/synapse/blob/master/lib/synapse.rb#L24-25)).

This means that any time synapse is restarted there is a period between when it starts and when the watcher first registers where the defaults are being used (and if there are no defaults you will return 503's from haproxy).

I can see why you would want an initial reconfigure so that any time you restart synapse you know any changes unrelated to backends will get picked up, but it seems like it should be smarter and wait until all watchers have checked in once before doing its initial reconfiguration.  I'm happy to submit a PR if this sounds reasonable.

jolynch:
@dcosson Yea this has been somewhat of an annoying bug to us as well. We mitigated it slightly at Yelp by reducing the amount of zookeeper connecting we do and using the state_file so that at the very least as soon as the watcher gets information we can quickly enable the backend over the stat socket, but this just shortens the window of sad.

How are you thinking about implementing this? The watchers all live in separate threads, but perhaps we can communicate back to the main thread a la `reconfigure`?

jolynch:
Now that I'm looking at it, the zookeeper watchers do setup before returning because they don't use a Thread.new in the start method (they rely on the zk-ruby gem's async callbacks to multithread themselves).

@dcosson I imagine you're using the ec2 tag watcher or some such?

jolynch:
Oops, didn't mean to close, sorry about that.

PR #208, opened 2016-08-24
comment line was breaking config comparison

davent:
Due to the auto-generated comment line included on the first line of the HAProxy, a restart of HAProxy will occur even when the configuration itself is identical.

Here is an example diff when no configuration has actually changed:
```
1c1
< # auto-generated by synapse at 2016-08-24 10:20:55 +0000
---
> # auto-generated by synapse at 2016-08-24 10:22:47 +0000
```
As the logic to decide whether or not the config is different is just a simple string comparison (within the write_config function) it will never return true.

### DIFF
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -866,7 +866,9 @@ module Synapse
 
       # if we write config files, lets do that and then possibly restart
       if @opts['do_writes']
-        write_config(new_config)
+        unless write_config(new_config)
+          @restart_required = false
+        end
         restart if @opts['do_reloads'] && @restart_required
       end
     end
@@ -1131,7 +1133,8 @@ module Synapse
         old_config = ""
       end
 
-      if old_config == new_config
+      if old_config.split("\n")[1..-1].join == new_config.split("\n")[1..-1].join
+        log.info "synapse: new config is identical to existing haproxy config file."
         return false
       else
         File.open(@opts['config_file_path'],'w') {|f| f.write(new_config)}
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -866,9 +866,7 @@ module Synapse
 
       # if we write config files, lets do that and then possibly restart
       if @opts['do_writes']
-        unless write_config(new_config)
-          @restart_required = false
-        end
+        @restart_required = false if write_config(new_config) == false
         restart if @opts['do_reloads'] && @restart_required
       end
     end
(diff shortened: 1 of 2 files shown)


jolynch:
@davent I don't see how this is a bug. As far as I know it is fine to write out the same HAProxy config because writing the config doesn't actually make anything restart. My understanding is that Synapse only tells HAProxy to restart when it detects (via the stats socket) that a restart is required, or when we first start Synapse.

When Synapse decided to restart HAProxy it usually logs a line saying why (a new backend or service), are there any such log lines before your suspect restarts?

davent:
Because ```write_config(new_config)``` always returns true then the following line:
```restart if @opts['do_reloads'] && @restart_required``` is execute every time. Even when no actual config changes have been made.

jolynch:
``write_config`` doesn't determine if HAProxy needs to be restarted. That's ``@restart_required`` which Synapse sets whenever it detects (via the stats socket or state file) that a restart is required. Are you running into the first time Synapse starts it restart HAProxy regardless of if the config has changed?

jolynch:
@davent I believe you're running into #78, but this PR isn't the right way to fix it. Now that I'm looking at the code though we really can't just rely on the config file changing to detect a reload as there is no guarantee that the HAProxy instance running is actually running with that config (imagine that synapse writes out the config and then right before it restarts HAProxy it get's restarted or killed). Our stat socket check at the beginning only guarantees that if we're missing backends we'll restart, but we can't detect unexpected backends or misconfigured backends.

Maybe we can get the pid of HAProxy out of the stat socket and then hash that and the config to determine a "HAProxy version" and if those match we can set ``@restart_required`` to false initially... What do you think?

jolynch:
@davent are you interested in getting the patch in shape or should I close this and continue in a separate PR.

I think the algorithm that works is:
1. Immediately after restarting HAProxy and first config generation, issue "show stat" to the stats socket and read out the "Pid" field.
2. Hash this value with the contents of the HAProxy config to produce a "Running HAProxy version", store that in the state file in restart, but read it in write_config (so we'll have to either add another state file or refactor the current one, shouldn't be hard)
3. Set ``@restart_required`` to be false in write_config iff the running haproxy version matches the last haproxy version

The state transition is (1) pending -> (2) config written -> (3) restarted -> (4) pid identified -> (5) pid + config hashed and written as a version. If we crash at any point before 5 we will identify a mismatch on Synapse startup (because we'll have the old PID) and guarantee a restart, which I beleive is a safe default.

There is some subtlety in when we read the version out, when we compute it, and how we manage the state transition of ``@restart_required`` (right now it's just a single latch, always F -> T and then T -> F only on restart, and that's changing ... so ... there is some complexity there), but I think the patch is pretty straightforward.

igor47:
oh using the pid is a great hack. i like the proposed approach.

jolynch:
@davent I'm going to close this and continue with the work in a separate PR. If you've got the patch feel free to resubmit here or in a different PR :-)

PR #209, opened 2016-09-10
Add splay startup jitter to haproxy startup

solarkennedy:
Sometimes all the synapses are restarted globally across a fleet, so I would like their startup to be jittered just like their normal `next_restart` is jittered during a normal haproxy restart. I believe this change will do that.


solarkennedy:
Bump? Every time we make a global change to our synapse config at Yelp we have a huge global blip, and I believe this is the cause/fix. cc @jolynch @evankrall

igor47:
this change looks good to me; was it tested?

solarkennedy:
Not any more than travisci.

PR #211, opened 2016-10-27
[haproxy] Avoid unnecessary HAProxy re-configuration.

noggi:
Currently, on every topology update event Synapse sends re-enable command to HAProxy for every live backend instance, even though that instance may already be `UP`. The number of such backend instances can be quite significant, so after every single topology update Synapse may send thousands of re-enable commands which are essentially noops.

The following changes were made to address the problem:
1. HAProxy command batching: commands are now sent in groups of 4 or less, which reduces
   overall number of syscalls per topology update event.
2. Re-enable command is now sent only if a given instance status is not `UP`.
3. Minor performance improvements in HAProxy stats parsing code (~35% time/cpu savings). 

NOTE: command batching in HAproxy was [added in version 1.4.0](http://git.haproxy.org/?p=haproxy.git;a=commit;h=9a42c0d771986913888dd2c3066dca059a850357) (early 2010)

### DIFF
--- a/lib/synapse/haproxy.rb
+++ b/lib/synapse/haproxy.rb
@@ -2,11 +2,14 @@ require 'fileutils'
 require 'json'
 require 'socket'
 require 'digest/sha1'
+require 'set'
 
 module Synapse
   class Haproxy
     include Logging
     attr_reader :opts
+
+    HAPROXY_CMD_BATCH_SIZE = 4
 
     # these come from the documentation for haproxy (1.5 and 1.6)
     # http://haproxy.1wt.eu/download/1.5/doc/configuration.txt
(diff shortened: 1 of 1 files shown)


igor47:
why does this PR have two copies of the first two commits?

lib/synapse/haproxy.rb
```diff
@@ -1085,36 +1092,44 @@ def update_backends_at(socket_file_path, watchers)
         watcher.backends.each do |backend|
           backend_name = construct_name(backend)
           if cur_backends[watcher.name].include? backend_name
-            enabled_backends[watcher.name] << backend_name
+            enabled_backends[watcher.name][backend_name] = 1
```
igor47: i'm having a lot of trouble understanding what `1` means here...

lib/synapse/version.rb
```diff
@@ -1,3 +1,3 @@
 module Synapse
-  VERSION = "0.13.8"
+  VERSION = "0.13.9"
```
igor47: don't bump the version as part of this PR. we should test this separately, and if it works then we can bump the version number as a separate change and then release the new version

lib/synapse/haproxy.rb
```diff
@@ -1060,20 +1064,23 @@ def update_backends_at(socket_file_path, watchers)
...
+
+        cur_backends[name] ||= {}
+        cur_backends[name][address] = state
       end
 
       # build a list of backends that should be enabled
       enabled_backends = {}
       watchers.each do |watcher|
-        enabled_backends[watcher.name] = []
+        enabled_backends[watcher.name] = {}
```
igor47: if you're turning this into a hash to speed up lookups, i would recommend using a Set instead

lib/synapse/haproxy.rb
```diff
@@ -8,6 +8,10 @@ class Haproxy
     include Logging
     attr_reader :opts
 
+    HAPROXY_CMD_BATCH_SIZE = 4
+
+    HAPROXY_BACKEND_STATE_UP = 'UP'
```
igor47: eh i kind of changed my mind about this. we hardcode a bunch of magic strings all through the code here, i don't think extracting this one constant improves the situation.

PR #216, opened 2016-11-25
Adding additional bind_options param to support ssl configuration

sjankowski-r7:
This PR adds an additional parameter to support the ssl configuration in haproxy. As specified in test and example config, those options could be: {ssl, crt, ciphers, accept-proxy, ca-file etc.}


udangel-r7:
bump

jolynch:
@sjankowski-r7 Thanks for the change. I'd like to refactor the bind section a bit more to allow primary binding to unix sockets rather than just host:ports, but I think this is reasonable along the way.

Also heads up that you can always inject custom bind lines using frontend_options as well (although not the primary bind line).

jolynch:
Also let me know if you want a version bump.

PR #218, opened 2017-01-22
Add Missing Quotation Mark in the Sample Config

femnad:
* The sample configuration file provided under the section `HAProxy shared HTTP Frontend` is not a valid YAML file because of a missing quotation mark. This patch adds the missing quotation mark so that the configuration is usable as it is.


jolynch:
Thanks :-)

PR #219, opened 2017-02-02
Add support for relative port overrides to the ec2 watcher.

udangel-r7:
haproxy supports relative port addresses, prefixed by a "+" or "-".
If this is set the server port is determined by adding the value to
the client's port.


udangel-r7:
hm now relative port overrides are no longer possible due the check being moved into base. Would it be ok to remove the check there?

PR #220, opened 2017-02-06
Nginx take two

jolynch:
This is my attempt at an nginx config generator based on the branch from #203

As I want to be able to force nginx to listen on the TCP port and not HAProxy, I had to make a minor change to the port contract there. Basically if port is null we don't insist it exists (and can bind to unix sockets instead).

I've included some very basic smoke tests and I'm working on carrying out some e2e tests for the nginx (TCP) -> HAProxy (unix) design for hitless proxies. @vulpine :-D

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    synapse (0.13.8)
+    synapse (0.14.0)
       aws-sdk (~> 1.39)
       docker-api (~> 1.7)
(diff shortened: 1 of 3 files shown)


jolynch:
Actually, I could probably do this as an "extension" of the synapse module using our dynamic plugin system. I hacked something up over the weekend and it seemed to work fine.

jolynch:
I'm going to close this because merges are strange. Will re-pr.

PR #223, opened 2017-03-10
Allow null bind port

jnb:
This is to allow us to support domain sockets.

### DIFF
--- a/spec/lib/synapse/haproxy_spec.rb
+++ b/spec/lib/synapse/haproxy_spec.rb
@@ -69,6 +69,15 @@ describe Synapse::ConfigGenerator::Haproxy do
     mockWatcher
   end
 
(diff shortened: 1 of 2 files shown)


spec/lib/synapse/haproxy_spec.rb
```diff
@@ -69,6 +69,15 @@ class MockWatcher; end;
     mockWatcher
   end
 
+  let(:mockwatcher_frontend_with_nil_port) do
+    mockWatcher = double(Synapse::ServiceWatcher)
+    allow(mockWatcher).to receive(:name).and_return('example_service6')
+    allow(mockWatcher).to receive(:config_for_generator).and_return({
+      'haproxy' => {'port' => nil, 'bind_address' => "127.0.0.3"}
```
jolynch: maybe make this a unix socket to really show what this is for?

PR #228, opened 2017-03-27
cache haproxy stanzas in generate_config

somic:
Two things here.

First, we toggle ```updated``` on a service watcher from its ```set_backends``` method if we detect a diff in backends.

Second, in haproxy generator, in ```generate_config``` method we use cached frontend and backend stanzas unless a given watcher has ```updated``` set to true.

cc @jolynch

### DIFF
--- a/lib/synapse/config_generator/haproxy.rb
+++ b/lib/synapse/config_generator/haproxy.rb
@@ -837,6 +837,11 @@ class Synapse::ConfigGenerator
       # a place to store the parsed haproxy config from each watcher
       @watcher_configs = {}
 
+      # a place to store generated frontend and backend stanzas
+      @frontends_cache = {}
+      @backends_cache = {}
+      @watcher_revisions = {}
+
       @state_file_path = @opts['state_file_path']
       @state_file_ttl = @opts.fetch('state_file_ttl', DEFAULT_STATE_FILE_TTL).to_i
     end
@@ -893,9 +898,6 @@ class Synapse::ConfigGenerator
       new_config = generate_base_config
       shared_frontend_lines = generate_shared_frontend
 
-      @frontends_cache ||= {}
-      @backends_cache ||= {}
-      @watcher_revisions ||= {}
       watchers.each do |watcher|
         watcher_config = watcher.config_for_generator[name]
         @watcher_configs[watcher.name] ||= parse_watcher_config(watcher)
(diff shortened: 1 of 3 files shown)


lib/synapse/config_generator/haproxy.rb
```diff
@@ -893,12 +893,22 @@ def generate_config(watchers)
...
-        new_config << generate_backend_stanza(watcher, @watcher_configs[watcher.name]['backend'])
+
+        if watcher.updated || @frontends_cache[watcher.name].nil?
+          @frontends_cache[watcher.name] = generate_frontend_stanza(watcher, @watcher_configs[watcher.name]['frontend'])
+        end
+        if watcher.updated || @backends_cache[watcher.name].nil?
+          @backends_cache[watcher.name] = generate_backend_stanza(watcher, @watcher_configs[watcher.name]['backend'])
+        end
+        new_config << @frontends_cache[watcher.name] << @backends_cache[watcher.name]
+        watcher.updated = false if watcher.updated
```
jolynch: I'm not a big fan of config_generator mutating service_watchers, they really need to be decoupled as much as possible IMO. Right now they have read only access to backends and config_for_generator ... but this would be adding complexity which is only present in haproxy (and might interfere with nginx or file_output).

Perhaps we can set this in the main loop after update_config is called on all config_generators? Or we can just have the config generator be 100% responsible for deciding to update or not (e.g. if backends haven't changed).

somic:
Good point, after you mentioned it, I also see how it's bad.

Adjusted for a watcher to keep track of its revision and haproxy will check if it's a revision it saw last time

lib/synapse/service_watcher/base.rb
```diff
@@ -6,13 +6,15 @@ class BaseWatcher
     include Synapse::Logging
 
     LEADER_WARN_INTERVAL = 30
+    MAX_REVISION = 99999999999
```
jolynch: I think this is a premature optimization, let's just count up and if it's an issue in the future we can fix it.

lib/synapse/config_generator/haproxy.rb
```diff
@@ -893,12 +893,24 @@ def generate_config(watchers)
       new_config = generate_base_config
       shared_frontend_lines = generate_shared_frontend
 
+      @frontends_cache ||= {}
```
jolynch: nit: let's put this up in initialize with the rest of setup.

somic:
@jolynch - I think I addressed both of your suggestions, tests pass, this is now ready.

PR #231, opened 2017-04-03
Adding tag_hash functionality to ec2tag watcher. 

microwavedboy:
Migrated to v2 of AWS-SDK
Updated rspec tests
Updated README.md to cover new functionality

This adds the ability to define a "tag_hash" key in the options block for the ec2tag watcher.

I had the use case of wanting to target only the intersection of two tags -- such as 
```
tag_hash:
  environment: staging
  some_service: installed
```

And the existing watcher functionality did not suffice. 

This is backwards compatible with existing synapse.yaml configuration files (and the tag_name/tag_value keys for ec2tag)

Cheers,
Alex


jolynch:
Can we somehow integrate this this solution with [label filters](https://github.com/airbnb/synapse/#filtering-service-nodes) instead? This way ec2tag filters nodes in a consistent way with the zookeeper watcher?

Perhaps ec2 instances could be tagged with a "synapse_labels" tag or something that get's emitted as labels from `discover_instances` ?

What do you think?

microwavedboy:
Having glanced at the label filters code -- I think I can probably rewrite
the code to have the ec2 watcher manipulate the 'labels' properties of each
backend object, and then rely on the base watcher's filter_labels behavior
to return the nodes I need.

PR #236, opened 2017-06-21
Propagate revision updates in ZookeeperDns watcher

noggi:
https://github.com/airbnb/synapse/pull/228 introduced stanza cache in haproxy generator, which relies on the revision counter to detect changes in configuration. This does not work properly with ZookeeperDns watcher which overrides `reconfigure!` method, so changes that happened in its child ZK watcher become invisible to the haproxy generator.

This PR allows for proper propagation of revision updates in ZookeeperDns.

@somic @jolynch @lucaluo @tuckerman @jimmyngo


jolynch:
Ah nice catch. Sorry about that breakage.

jolynch:
Released as 0.14.6 (should be on rubygems by now)

PR #237, opened 2017-07-07
Bump revision in DNS watcher of ZK DNS

noggi:
My previous PR (https://github.com/airbnb/synapse/pull/236) only fixed the ZK 
part of ZK DNS watcher. As the result, only ZK updates were handled correctly.

This change fixes the DNS part as well, so IP address updates will now also
trigger haproxy reconfiguration.

@jimmyngo @tuckerman @jolynch

### DIFF
--- a/lib/synapse/service_watcher/zookeeper_dns.rb
+++ b/lib/synapse/service_watcher/zookeeper_dns.rb
@@ -51,8 +51,9 @@ class Synapse::ServiceWatcher
       # Overrides the discovery_servers method on the parent class
       attr_accessor :discovery_servers
 
-      def initialize(opts={}, synapse, message_queue)
+      def initialize(opts={}, parent=nil, synapse, message_queue)
         @message_queue = message_queue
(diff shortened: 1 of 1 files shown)


tuckerman:
LGTM

jimmyngo:
LGTM!

lib/synapse/service_watcher/zookeeper_dns.rb
```diff
@@ -51,8 +51,9 @@ class Dns < Synapse::ServiceWatcher::DnsWatcher
       # Overrides the discovery_servers method on the parent class
       attr_accessor :discovery_servers
 
-      def initialize(opts={}, synapse, message_queue)
+      def initialize(opts={}, parent=nil, synapse, message_queue)
```
tuckerman: is this instantiated anywhere other than your change below? if not, why the need for the default arg?

jolynch:
Thanks for the fix :-)

Issue #238, opened 2017-07-10
Synapse and HAProxy seamless reloads


StephanErb:
Excuse me for abusing this issue tracker as way to ask a user question:

I am wondering if anyone has experience running synapse with the HAProxy seamless restart settings `server-state-file`, `load-server-state-from-file global` and the extended  reload command 
```
socat haproxy.sock - <<< 'show servers state' > haproxy.state
systemctl haproxy reload
```
 (For details, see https://github.com/haproxy/haproxy/blob/master/examples/seamless_reload.txt)

Are there any known issues using this features, given that synapse will re-configure haproxy immediately after re-launching? 

Beyond a few warnings due to different service ordering it appears to be working. However, I have recently noticed a few backends enter a state where they never exit the `DOWN` state even though they can be reached from the proxy host.  Before I dig deeper into HAProxy I would like to confirm that I am not doing something stupid here.
 
```
Jul 10 07:57:07 proxy1 haproxy-systemd-wrapper[5649]: haproxy-systemd-wrapper: re-executing on SIGUSR2.
Jul 10 07:57:07 proxy1 haproxy-systemd-wrapper[5649]: haproxy-systemd-wrapper: executing /usr/sbin/haproxy -f /var/lib/synapse/haproxy.cfg -p /run/haproxy.pid -Ds -sf 9990
Jul 10 07:57:07 proxy1 haproxy-systemd-wrapper[5649]: [WARNING] 190/075707 (31569) : can't find server '0_mesos-node1.example.org:31724' with id '2' in backend with id '4' or name 'foo2-service.example.orgg'
Jul 10 07:57:07 proxy1 haproxy[31569]: can't find server '0_mesos-node1.example.org:31724' with id '2' in backend with id '4' or name 'foo2-service.example.orgg'
Jul 10 07:57:07 proxy1 haproxy[31569]: can't find server '0_mesos-node1.example.org:31724' with id '2' in backend with id '4' or name 'foo2-service.example.orgg'
Jul 10 07:57:07 proxy1 haproxy-systemd-wrapper[5649]: [WARNING] 190/075707 (31569) : In backend 'foo1-service.example.org' (id: '5'): server ID mismatch: from server state file: '1', from running config 2
Jul 10 07:57:07 proxy1 haproxy[31569]: In backend 'foo1-service.example.org' (id: 5): server ID mismatch: from server state file: '1', from running config 2
[...]

PR #239, opened 2017-07-11
Make shuffle reproducible

jolynch:
This way we still get different orders on different client machines,
preventing long lived connections from all hitting the first or second
server, but also for reloads we get a consistent ordering that does not
invalidate the HAProxy state file (which assumes your backends don't
change order).

### DIFF
--- a/lib/synapse/config_generator/haproxy.rb
+++ b/lib/synapse/config_generator/haproxy.rb
@@ -854,6 +854,9 @@ class Synapse::ConfigGenerator
       # Map of backend names -> hash of HAProxy server names -> puids
       # (server->id aka "name") to their proxy unique id (server->puid aka "id")
       @server_id_map = Hash.new{|h,k| h[k] = {}}
+      # Map of backend names -> hash of HAProxy server puids -> names
+      # (server->puid aka "id") to their name (server->id aka "name")
+      @id_server_map = Hash.new{|h,k| h[k] = {}}
 
     end
 
@@ -1049,13 +1052,12 @@ class Synapse::ConfigGenerator
       # The ordering here is important.  First we add all the backends in the
       # disabled state...
       seen.fetch(watcher.name, []).each do |backend_name, backend|
-        backends[backend_name] = backend.merge(
-          'enabled' => false,
-        )
+        backends[backend_name] = backend.merge('enabled' => false)
         # We remember the haproxy_server_id from a previous reload here.
         # Note though that if live servers below define haproxy_server_id
         # that overrides the remembered value
-        @server_id_map[watcher.name][backend_name] ||= backends[backend_name]['haproxy_server_id']
+        @server_id_map[watcher.name][backend_name] ||= backends[backend_name]['haproxy_server_id'].to_i
+        @id_server_map[watcher.name][@server_id_map[watcher.name][backend_name]] = backend_name
       end
 
       # ... and then we overwite any backends that the watchers know about,
@@ -1076,23 +1078,26 @@ class Synapse::ConfigGenerator
         backends[backend_name] = backend.merge('enabled' => true)
 
         # If the the registry defines the haproxy_server_id that must be preferred.
+        # Note that the order here is important, because if haproxy_server_options
+        # does define an id, then we will write that out below, so that must be what
+        # is in the id_map as well.
+        @server_id_map[watcher.name][backend_name] = backend['haproxy_server_id'].to_i if backend['haproxy_server_id']
         server_opts = (backend['haproxy_server_options'] || 'no match').split(' ')
         @server_id_map[watcher.name][backend_name] = server_opts[server_opts.index('id') + 1].to_i if server_opts.include?('id')
-        # This is going to be an issue ... and haproxy will fail if this happens ... oh well
-        @server_id_map[watcher.name][backend_name] = backend['haproxy_server_id'] if backend['haproxy_server_id']
+        @id_server_map[watcher.name][@server_id_map[watcher.name][backend_name]] = backend_name
       end
 
       # Now that we know the maximum possible existing haproxy_server_id for
       # this backend, we can set any that don't exist yet.
       watcher.backends.each do |backend|
         backend_name = construct_name(backend)
-        @server_id_map[watcher.name][backend_name] ||= (
-          ((@server_id_map[watcher.name].values + [0]).compact.max) % @max_server_id + 1
-        )
+        @server_id_map[watcher.name][backend_name] ||= find_next_id(watcher.name, backend_name)
+        @id_server_map[watcher.name][@server_id_map[watcher.name][backend_name]] = backend_name
       end
       # Remove any servers that don't exist anymore from the server_id_map
       # to control memory growth
-      @server_id_map[watcher.name].select! { |server_name| backends.has_key?(server_name)}
+      @server_id_map[watcher.name].keep_if { |server_name| backends.has_key?(server_name) }
+      @id_server_map[watcher.name].keep_if { |_,  server_name| @server_id_map[watcher.name][server_name] }
 
       if watcher.backends.empty?
         log.debug "synapse: no backends found for watcher #{watcher.name}"
@@ -1136,6 +1141,23 @@ class Synapse::ConfigGenerator
           b = "#{b} disabled" unless backend['enabled']
           b }
       ]
+    end
+
+    def find_next_id(watcher_name, backend_name)
+      probe = nil
+      if @server_id_map[watcher_name].size >= @max_server_id
+        log.error "synapse: ran out of server ids for #{watcher_name}, if you need more increase the max_server_id option"
+        return probe
+      end
+
+      max_existing_id = @id_server_map[watcher_name].keys()[-1] || 0
+      probe = max_existing_id % @max_server_id + 1
+
+      while @id_server_map[watcher_name].include?(probe)
+        probe = (probe % @max_server_id) + 1
+      end
+
+      probe
     end
 
     def talk_to_socket(socket_file_path, command)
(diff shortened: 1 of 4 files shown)


jolynch:
@minkovich this will help a lot with conflicts during synapse reloads/restarts. I asked on the HAProxy mailing list and Baptiste pointed me towards the server id option, which we can totally use to make it work 100% I believe (even if the order does shift around).

That change is a bit ... more involved though, so I was thinking of doing that change separately from this one. Or I can group them together, which do you prefer?

minkovich:
@jolynch That makes sense. I'm in favor of smaller changes.

jolynch:
@minkovich I oopsed and fixed the whole problem. I believe that the solution in this pull request should make things work 100% with the state file since we remember the id that we assigned backends last go around.

I'm slightly concerned with how this potentially conflicts between haproxy_server_options and haproxy_server_id ... but ... I think it's fine.

This should fix many of the issues brought up in #238 (not the DOWN part, but it should make the HAProxy state file actually work) . @StephanErb heads up

lib/synapse/config_generator/haproxy.rb
```diff
@@ -1038,7 +1049,13 @@ def generate_backend_stanza(watcher, config)
       # The ordering here is important.  First we add all the backends in the
       # disabled state...
       seen.fetch(watcher.name, []).each do |backend_name, backend|
-        backends[backend_name] = backend.merge('enabled' => false)
+        backends[backend_name] = backend.merge(
```
minkovich: Convert back to single line for consistency?

lib/synapse/config_generator/haproxy.rb
```diff
@@ -1055,9 +1072,28 @@ def generate_backend_stanza(watcher, config)
             @restart_required = true
           end
         end
+
         backends[backend_name] = backend.merge('enabled' => true)
+
+        # If the the registry defines the haproxy_server_id that must be preferred.
+        server_opts = (backend['haproxy_server_options'] || 'no match').split(' ')
+        @server_id_map[watcher.name][backend_name] = server_opts[server_opts.index('id') + 1].to_i if server_opts.include?('id')
+        # This is going to be an issue ... and haproxy will fail if this happens ... oh well
```
minkovich: Can you elaborate on this comment by explaining the conflict?

lib/synapse/config_generator/haproxy.rb
```diff
@@ -1055,9 +1072,28 @@ def generate_backend_stanza(watcher, config)
...
+        @server_id_map[watcher.name][backend_name] = backend['haproxy_server_id'] if backend['haproxy_server_id']
       end
 
+      # Now that we know the maximum possible existing haproxy_server_id for
+      # this backend, we can set any that don't exist yet.
+      watcher.backends.each do |backend|
+        backend_name = construct_name(backend)
+        @server_id_map[watcher.name][backend_name] ||= (
+          ((@server_id_map[watcher.name].values + [0]).compact.max) % @max_server_id + 1
```
minkovich: This isn't very efficient and can lead to a problem with three backends (a,b,c) but two set haproxy_server_options with id = 1 (for a) and the other id = max_server_id (for b). This loop could set id (for c) to also be 1 causing a conflict. This situation is nearly impossible but should still be addressed. A reverse index id_server_map[watcher.name][id] -> backend_name might fix this.

lib/synapse/config_generator/haproxy.rb
```diff
@@ -1055,9 +1072,28 @@ def generate_backend_stanza(watcher, config)
...
+        )
+      end
+      # Remove any servers that don't exist anymore from the server_id_map
+      # to control memory growth
+      @server_id_map[watcher.name].select! { |server_name| backends.has_key?(server_name)}
```
minkovich: Just a personal preference but keep_if is a bit easier to read

spec/lib/synapse/haproxy_spec.rb
```diff
@@ -21,7 +21,18 @@ class MockWatcher; end;
   let(:mockwatcher_with_server_options) do
     mockWatcher = double(Synapse::ServiceWatcher)
     allow(mockWatcher).to receive(:name).and_return('example_service2')
-    backends = [{ 'host' => 'somehost', 'port' => 5555, 'haproxy_server_options' => 'backup'}]
+    backends = [{ 'host' => 'somehost', 'port' => 5555, 'haproxy_server_options' => 'id 12 backup'}]
```
minkovich: Could you add a test with several backends and one setting 'haproxy_server_options' => 'id <max_server_id>'

lib/synapse/config_generator/haproxy.rb
jolynch: done

lib/synapse/config_generator/haproxy.rb
jolynch: thinking about this more, as long as we always choose the ``haproxy_server_options`` value over the ``haproxy_server_id`` this won't conflict, we'll just always use the haproxy_server_id value.

I was referring to the conflict when a server defines two ids (we do it and then later the server_options do it). But as long as we always default to the id in ``haproxy_server_options`` I think we're good.

lib/synapse/config_generator/haproxy.rb
jolynch: Sure, changed.

spec/lib/synapse/haproxy_spec.rb
jolynch: On it.

lib/synapse/config_generator/haproxy.rb
jolynch: Hm, I'm less concerned about someone setting manual ids and more concerned with holes created by a churning server (e.g. srv1 has id 1, and then you have a container start and stop 65k times until it loops over), but I see what you mean.

I do wish that ruby had a better stdlib (heap or sortedset without rbtree), but tbh I don't think that having an O(n) probe here is _that_ bad. N is guaranteed to be reasonably small (65k) and we only end up doing it when we have a new server. Sorted set does seem like exactly what I want ... but ... it's slow ... I'll figure something out.

spec/lib/synapse/haproxy_spec.rb
jolynch: I refactored the haproxy_server_id test to test all three ways of specifying ids (no specification, one with haproxy_server_id and one with haproxy_server_options). Test failed before I changed the algorithm for picking ids (duplicate id of 1) and then passed after.

jolynch:
@minkovich I like the reverse map because it also paves a way forward for supporting minimal reload synapse. Right now it doesn't quite work, but I think it would be relatively easy to pre-allocate a bunch of servers, stop writing out the names, and just go by id. We can change the socket commands to use the "#puid" variant (instead of "name" and we're in business.

So basically we'd pre-allocate a bunch of servers and then dynamically ``set server <backend>/<#puid> addr`` and we never have to restart ever again heh.

lib/synapse/config_generator/haproxy.rb
```diff
@@ -1138,6 +1143,23 @@ def generate_backend_stanza(watcher, config)
...
     end
 
+    def find_next_id(watcher_name, backend_name)
+      probe = nil
+      if @server_id_map[watcher_name].size >= @max_server_id
+        log.error "synapse: ran out of server ids for #{watcher_name}, if you need more increase the max_server_id option"
+        return probe
+      end
+
+      max_existing_id = @id_server_map[watcher_name].keys()[-1] || 0
```
minkovich: I think you need to call max() on keys()?

jolynch: indeed

minkovich:
@jolynch Is the plan still to merge this PR then make the larger refactor? We are blocked from rolling out some of the new feature of HAProxy (#238).

jolynch:
I'm going to merge this tomorrow, just getting back from a trip.

jolynch:
@minkovich @StephanErb can you guys build from a sha or do you need a release? I wanted to roll it out here first and make sure it works before cutting a release but I can cut one if you need it.

StephanErb:
Thanks a lot for the patch! 

Testing this one out before cutting a release sounds fine to me.

darnaut:
We recently upgraded synapse and now haproxy cannot be loaded:

```
2017-09-14T03:02:21.040206+00:00 localhost.localdomain user.notice [ALERT] - 256/030220 (20806) : parsing [/etc/haproxy/haproxy.cfg:10313] : 'server i-xxx:yyyy' : 'id' : custom id 2 already used at /etc/haproxy/haproxy.cfg:10292 ('server i-zzz:yyyy')
```

jolynch:
@darnaut can you supply the HAProxy backend that is affected (server lines) and the synapse watcher configuration for that backend? Thanks :-)

Also you might want to make an issue report instead of commenting on this PR, up to you.

darnaut:
Turned out I had to rollback a few more commits for downgrade to work.

Would appreciate if new features such as time came disabled by default so rollout can be separated from upgrades.

jolynch:
@darnaut this change should be fully backwards compatible (fwiw it was with our config). Previously Synapse was not managing server ids and the change should prefer ids provided in server_options (the only way previously to provide a server id). If you provide the info from ^^ I can help you debug what's going on :-)

darnaut:
@jolynch Sorry, I'm confused. You are saying this is not enabled by default?

jolynch:
@darnaut It is enabled by default, but it should be fully backwards compatible with previous configs 
(specifically if someone was previously setting server ids, this change should respect those settings over the generated ones).

This change fixes a bug (that the ordering of servers changes per reload and that we don't give HAProxy servers consistent ids through reloads). It is should not provide breaking functionality. If Synapse was outside a 0.X release this would be a "minor" version bump. If it broke your config then clearly I made a mistake, but having some more details will help fix it so that future folks don't have an issue :-)

darnaut:
@jolynch Thanks for the explanation. Well, it clearly broke for us. Also our config has several hundred services, so will take a while to isolate that and pinpoint exactly what is causing the config to break. Also, and more pressing, there is some other problem in master branch that causes several haproxy instances after synapse is restarted due to disabled backends disappearing from haproxy config.

jolynch:
@darnaut The line that you posted has the line number of the HAProxy config that is invalid, if you can include that server line's backend (all the servers in the enclosing backend) and the relevant synapse service watcher config, I can try to figure out what went wrong (possibly need the synapse state file's entry for that service watcher as well but we'll see).

PR #242, opened 2017-09-14
[synapse] fix bug in generate_backend_stanze

lap1817:
to: @jolynch 

## what
A bug was introduced into the generate_Backend_stanza. @state_cache.backends is expecting watcher.name as the input argument, but it got the watcher object.
With the bug, when Synapse generates the haproxy config, it only sees the "enabled" services as reported by the watchers, but does not see all known backends in the state cache (because search by 'watch', instead of watcher.name, in the state cache failed silently). This increases number of haproxy restarts when Synapse is restarted. 

## fix
Pass the correct argument to @state_cache.backends

## test
Added rspec test to cover this bug.

### DIFF
--- a/lib/synapse/config_generator/haproxy.rb
+++ b/lib/synapse/config_generator/haproxy.rb
@@ -1068,7 +1068,7 @@ class Synapse::ConfigGenerator
 
       # The ordering here is important.  First we add all the backends in the
       # disabled state...
-      @state_cache.backends(watcher).each do |backend_name, backend|
+      @state_cache.backends(watcher.name).each do |backend_name, backend|
         backends[backend_name] = backend.merge('enabled' => false)
         # We remember the haproxy_server_id from a previous reload here.
         # Note though that if live servers below define haproxy_server_id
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,7 +2,6 @@ language: ruby
 cache: bundler
 sudo: false
 rvm:
-  - 1.9.3-p551
```
jolynch: Did you mean to remove this? I think some companies still run this version (I know ...)

lap1817:
@jolynch , I am also removing Travis build on rvm 1.9.3-p551. Travis CI is having odd failures. I made a testing PR which only changed the README file, but still got the same CI build failure on rvm 1.9.30p551. https://github.com/airbnb/synapse/pull/243

The error is likely due to the old version of bundler (1.7.6) being used (other builds use 1.15.4). I found a discussion thread related to this: https://github.com/bundler/bundler/issues/3558. Seems that the issue is fixed in bundler version after 1.7.15. 

I tried to enforce upgrading bundle in travis CI before bundle install. But it didn't work either. 

So I am removing the build on rvm 1.9.3.

PTAL

.travis.yml
lap1817: just replied below in the comments. sounds good to you?

jolynch:
@lap1817 I just merged a workaround (upgrade bundler before_install), can you pull master and keep 1.9?

lap1817:
sure. doing it. odd that it didn't work for me

jolynch:
Thanks @lap1817, I'll cut a bug fix release in a sec.

lap1817:
thank you @jolynch

jolynch:
0.15.1 cut with the fix, cheers.

PR #244, opened 2017-09-15
Try upgrading bundler on travis

jolynch:
THis is a test to see if it fixes the 1.9 issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,6 @@ rvm:
   - 2.1.10
   - 2.2.5
   - 2.3.1
+
+before_install:
+  - gem install bundler


jolynch:
Strange, not sure why this works (bundler is the same version before and after), but it does so ... strange

jolynch:
Oh, yea, 1.7 vs 1.15, alrighty then.

greysteil:
FYI, the underlying issue here is that Bundler only introduced `required_ruby_version` in Bundler 1.11. As a result, with Bundler 1.7 you could get a bundle that didn't work with you Ruby version (e.g., a ActiveSupport 5.1.4 on Ruby 1.9).

PR #245, opened 2017-10-07
Support embedded ruby (ERB) in config files

darnaut:
Allows for string interpolation and other Ruby syntax in config files. Useful to dynamically build services based on local environment.

### DIFF
--- a/spec/bin/synapse_spec.rb
+++ b/spec/bin/synapse_spec.rb
@@ -1,0 +1,56 @@
+require 'spec_helper'
+require 'tempfile'
+
+describe 'parseconfig' do
+  it 'parses a templated config' do
+    allow_any_instance_of(Synapse::Synapse).to receive(:run)
+    expect(Synapse::Synapse).to receive(:new).
+      with(hash_including(
+       {"services" =>
+          {"test" =>
+            {"default_servers" =>
+              [{"name" => "default1", "host" => "localhost", "port" => 8080}],
+             "discovery" =>
+              {"method" => "zookeeper",
+               "path" => "/airbnb/service/logging/event_collector",
+               "hosts" => ["localhost:2181"],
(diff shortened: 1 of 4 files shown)


spec/bin/synapse_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'spec_helper'
+
+describe 'parseconfig' do
+  it 'parses a templated config' do
+    allow_any_instance_of(Synapse::Synapse).to receive(:run).and_return(nil)
```
weeyum: nit: stubbing a call will already make it return nil. we can remove `and_return(nil)`

spec/bin/synapse_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+           "config_file_path" => "/etc/haproxy/haproxy.cfg",
+           "do_writes" => false,
+           "do_reloads" => false,
+           "do_socket" => false,
+           "global" => ["global_test_option"],
+           "defaults" => ["default_test_option"]},
+         "file_output" => {"output_directory" => "/tmp/synapse_file_output_test"}}
+      )).and_call_original
+    ENV['SYNAPSE_CONFIG'] = "#{File.dirname(__FILE__)}/../support/minimum.conf.yaml"
+    ENV['SYNAPSE_CONFIG_VALUE'] = 'config value'
```
weeyum: reset ENV variables after the test, otherwise they will not be cleaned up.
2 options:
1) set original ENV var to a variable, ie: `cached_env_synapse_config`, assign it back after
2) stub the call to ENV, ie: `allow(ENV).to receive(:[]).with('SYNAPSE_CONFIG').and_return(....)`

noggi:
LGTM

weeyum:
LGTM

PR #246, opened 2017-11-10
add generator_config_path config option

bsherrod:
## Summary

This PR adds a `generator_config_path` option that can be used to determine where the zookeeper discovery method will look in zookeeper for the generator config.    This PR also adds a "disabled" option to prevent generator configs from being read from zookeeper.

Prior to this patch, the `path` option is used, and if the `path` node in zookeeper has data, that data always overrides any other generator config.

However, there are cases where you want to have multiple generator configs that share the same instances; for example, for az-aware routing.      In order to safely roll out this feature, you may also swant to disable the zookeeper-based loading of generator config information.

### Example:

For az-aware routing, you might implement this with two smartstack services, as follows:

```
  "services": {
    "sampleapp": {
      "discovery": {
        "method": "zookeeper",
        "path": "/production/services/sampleapp/services"
      },
      "haproxy": {
        "port": 11001,
        "server_options": "check inter 10s rise 8 fall 9",
        "frontend": [
          "acl use_local_backend nbsrv(sampleapp-local-az) ge 1",
          "use_backend sampleapp-local-az if use_local_backend",
          "bind ::1:11001"
        ],
        "listen": [
          "mode http",
          "option httpchk GET /health",
          "cookie SRVID insert indirect nocache",
          "option httplog"
        ]
      }
    },
    "sampleapp-local-az": {
      "discovery": {
        "label_filter": {
          "condition": "equals",
          "label": "az",
          "value": "us-east-1b"
        },
        "method": "zookeeper",
        "path": "/production/services/sampleapp/services"
      },
      "haproxy": {
        "server_options": "check inter 10s rise 8 fall 9",
        "frontend": [
          "acl use_local_backend nbsrv(sampleapp-local-az) ge 1",
          "use_backend sampleapp-local-az if use_local_backend",
          "bind ::1:11001"
        ],
        "listen": [
          "mode http",
          "option httpchk GET /health",
          "cookie SRVID insert indirect nocache",
          "option httplog"
        ]
      }
    }
  }
```

In this example, `sampleapp` and `sampleap-local-az` share the same path for instance discovery in zookeeper, but require different generator configs (different `haproxy` sections).     This isn't possible to express with the current zookeeper-based generator config discovery.     With this PR, this can be done as:

```
  "services": {
    "sampleapp": {
      "discovery": {
        "method": "zookeeper",
        "path": "/production/services/sampleapp/services",
        "generator_config_path": "/production/services/sampleapp/services"
      }
    },
    "sampleapp-local-az": {
      "discovery": {
        "label_filter": {
          "condition": "equals",
          "label": "az",
          "value": "us-east-1b"
        },
        "method": "zookeeper",
        "path": "/production/services/sampleapp/services",
        "generator_config_path": "/production/services/sampleapp-local-az/services"
      }
    }
  }
```

### DIFF
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -157,11 +157,30 @@ class Synapse::ServiceWatcher
         end
       end
 
-      node = @zk.get(@discovery['path'], :watch => true)
-      begin
-        new_config_for_generator = parse_service_config(node.first)
-      rescue StandardError => e
-        log.error "synapse: invalid config data in ZK node at #{@discovery['path']}: #{e}"
+      # support for a separate 'generator_config_path' key, for reading the
+      # generator config block, that may be different from the 'path' key where
+      # we discover service instances. if generator_config_path is present and
+      # the value is "disabled", then skip all zk-based discovery of the
+      # generator config (and use the values from the local config.json
+      # instead).
+      case @discovery.fetch('generator_config_path', nil)
+      when 'disabled'
+        discovery_key = nil
+      when nil
+        discovery_key = 'path'
+      else
+        discovery_key = 'generator_config_path'
+      end
+
+      if discovery_key
+        node = @zk.get(@discovery[discovery_key], :watch => true)
+        begin
+          new_config_for_generator = parse_service_config(node.first)
+        rescue StandardError => e
+          log.error "synapse: invalid config data in ZK node at #{@discovery[discovery_key]}: #{e}"
+          new_config_for_generator = {}
+        end
+      else
         new_config_for_generator = {}
       end
 
--- a/README.md
+++ b/README.md
@@ -230,7 +230,8 @@ It takes the following mandatory arguments:
 
 The watcher assumes that each node under `path` represents a service server.
 
-The watcher assumes that the data (if any) retrieved at znode `path` is a hash, where each key is named by a valid `config_generator` (e.g. `haproxy`) and the value is a hash that configs the generator.
+The watcher assumes that the data (if any) retrieved at znode `path` is a hash, where each key is named by a valid `config_generator` (e.g. `haproxy`) and the value is a hash that configs the generator.   Alternatively, if a `generator_config_path` argument is specified, the watcher will attempt to read generator config from that znode instead.
+If `generator_config_path` has the value `disabled`, then generator config will not be read from zookeeper at all.
 
 The following arguments are optional:
 
--- a/spec/lib/synapse/service_watcher_zookeeper_spec.rb
+++ b/spec/lib/synapse/service_watcher_zookeeper_spec.rb
@@ -124,6 +124,44 @@ describe Synapse::ServiceWatcher::ZookeeperWatcher do
       expect(subject).to receive(:set_backends).with([],{})
       subject.send(:watcher_callback).call
     end
+
+    context "when generator_config_path is defined" do
+      let(:discovery) { { 'method' => 'zookeeper', 'hosts' => 'somehost', 'path' => 'some/path', 'generator_config_path' => 'some/other/path' } }
+
+      it 'reads from generator_config_path znode' do
+        expect(subject).to receive(:watch)
+        expect(subject).to receive(:discover).and_call_original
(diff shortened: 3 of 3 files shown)


lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -157,12 +157,32 @@ def discover
...
-        log.error "synapse: invalid config data in ZK node at #{@discovery['path']}: #{e}"
-        new_config_for_generator = {}
+      # support for a separate 'generator_config_path' key, for reading the
+      # generator config block, that may be different from the 'path' key where
+      # we discover service instances. if generator_config_path is present and
+      # the value is "disabled", then skip all zk-based discovery of the
+      # generator config (and use the values from the local config.json
+      # instead).
+      if @discovery.has_key?('generator_config_path')
+        if @discovery['generator_config_path'] != 'disabled'
```
igor47: using two conditionals won't work; if the key is present, but set to `'disabled'`, you won't fall back to the `else` condition below, and will end up with neither the config from zk nor the config from the config file. for clarity, i would declare a variable like so:

```ruby
generator_config_path = @discovery['generator_config_path']
if generator_config_path && generator_config_path != 'disabled'
  stuff
else
  other stuff
end
```
use_config_from_zk = @discovery.has_key?('genreator_confi

bsherrod: That's the desired behavior:  If the key is present and set to "disabled", it should disable any zk based discovery, including the "path" block below.

Otherwise there is no way to disable the "path" lookup.

bsherrod:
Hi Igor!   Thank you for the quick review.    The change to synapse to support fetching configs from ZK was already introduced in https://github.com/airbnb/synapse/pull/241.   This PR makes that logic more flexible.

Both that PR and this one are being made by Airbnb engineers (I'm currently at Airbnb).

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -157,12 +157,32 @@ def discover
         end
       end
 
-      node = @zk.get(@discovery['path'], :watch => true)
```
darnaut: This whole block of code got duplicated without need. Just use a variable to hold the path and this block can remain about the same.

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -157,12 +157,30 @@ def discover
...
-      rescue StandardError => e
-        log.error "synapse: invalid config data in ZK node at #{@discovery['path']}: #{e}"
-        new_config_for_generator = {}
+      # support for a separate 'generator_config_path' key, for reading the
+      # generator config block, that may be different from the 'path' key where
+      # we discover service instances. if generator_config_path is present and
+      # the value is "disabled", then skip all zk-based discovery of the
+      # generator config (and use the values from the local config.json
+      # instead).
+      if @discovery.has_key?('generator_config_path')
```
darnaut: Would suggest rewriting this with a case statement:

```
case @discovery.fetch('generator_config_path', nil)
when 'disabled'
  discovery_key = nil
when nil
  discovery_key = 'path'
else
  discovery_key = 'generator_config_path'
end
```

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -157,12 +157,30 @@ def discover
...
+        if @discovery['generator_config_path'] == 'disabled'
+          discovery_key = nil
+        else
+          discovery_key = 'generator_config_path'
+        end
+      else
+        discovery_key = 'path'
+      end
+
+      unless discovery_key.nil?
```
darnaut: Use `if discovery_key`

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -157,12 +157,30 @@ def discover
...
+        node = @zk.get(@discovery[discovery_key], :watch => true)
+        begin
+          new_config_for_generator = parse_service_config(node.first)
+        rescue StandardError => e
+          log.error "synapse: invalid config data in ZK node at #{@discovery[discovery_key]}: #{e}"
+          new_config_for_generator = {}
+        end
```
darnaut: `new_config_for_generator` is not initialized if discovery_key is nil.

darnaut:
LGTM

PR #248, opened 2018-01-13
[Synapse] atomic config file update

lap1817:
reviewer:
@jolynch @liangg

This PR is to enable atomic Haproxy config file update.

The issue that we observed is that when Synapse restarts Haproxy in a high frequency (e.g. every 2 seconds), the haproxy config file may get partially over-written while a Haproxy process is still loading the file, especially for services that have high number of dependencies and thus have large haproxy config (1MB+). In some cases, the haproxy process corrupted because of that.


jolynch:
Thanks @lap1817

PR #250, opened 2018-03-16
[Synapse] fix synapse panic in short disconnection

lap1817:
Summary

This PR makes synapse watcher to keep alive as long as the session has not expired. The change avoids unnecessary Synapse restart when there is only short network disconnection between watcher and the zk cluster.

Reviewers
@darnaut @alexism

### DIFF
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -47,7 +47,9 @@ class Synapse::ServiceWatcher
     def ping?
       # @zk being nil implies no session *or* a lost session, do not remove
       # the check on @zk being truthy
-      @zk && @zk.connected?
+      # if the client is in any of the three states: associating, connecting, connected
+      # we consider it alive. this can avoid synapse restart on short network dis-connection
+      @zk && (@zk.associating? || @zk.connecting? || @zk.connected?)
     end
 
     private
--- a/spec/lib/synapse/service_watcher_zookeeper_spec.rb
+++ b/spec/lib/synapse/service_watcher_zookeeper_spec.rb
@@ -125,6 +125,15 @@ describe Synapse::ServiceWatcher::ZookeeperWatcher do
       subject.send(:watcher_callback).call
     end
 
+    it 'responds fail to ping? when the client is not in any of the connected/connecting/associatin state' do
+      expect(mock_zk).to receive(:associating?).and_return(false)
+      expect(mock_zk).to receive(:connecting?).and_return(false)
+      expect(mock_zk).to receive(:connected?).and_return(false)
+
+      subject.instance_variable_set('@zk', mock_zk)
+      expect(subject.ping?).to be false
+    end
+    
     context "when generator_config_path is defined" do
       let(:discovery) { { 'method' => 'zookeeper', 'hosts' => 'somehost', 'path' => 'some/path', 'generator_config_path' => 'some/other/path' } }
 


lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -47,7 +47,9 @@ def stop
     def ping?
       # @zk being nil implies no session *or* a lost session, do not remove
       # the check on @zk being truthy
-      @zk && @zk.connected?
+      # check expired_session? instead of connected?
+      # so that in case of short dis-connection, the session may be resued after connected
+      @zk && @zk.expired_session?
```
jolynch: Hm, I'm not sure about this, my understanding of `expired_session` is it only returns true if the session was expired (in which case we'd want the inverse right?

Given that we handle session expiry with a callback https://github.com/airbnb/synapse/blob/36457bb5b15a3be9fd8fb38b149da69f331ee054/lib/synapse/service_watcher/zookeeper.rb#L264-L267
perhaps this should be `&& (@zk.connected? || @zk.connecting?)`

alexism: regarding 

> perhaps this should be && (@zk.connected? || @zk.connecting?)

You're correct. Stating the states that are valid would be better. I would add the `associating` state as well for thoroughness.
So:
```
@zk && (@zk.associating? || @zk.connecting? || @zk.connected?)
```

lap1817: @jolynch , please correct me if I am wrong on this: the 'reconnecting' stat is set by zk.reconnect ,which is only triggered by zk.on_expired_session callback. In the network disconnected case, the client won't get the session_expire event, and so that the call to ping will still return fail.

The problem we observe is that, watcher got ping call in a short network disconnection. If my understanding above is correct, `&& (@zk.connected? || @zk.connecting?)` won't help our case either?

jolynch: @lap1817 Hm, yea I see what you mean, although if we have zk session loss we're going to fail pings because of the [callback](https://github.com/airbnb/synapse/blob/36457bb5b15a3be9fd8fb38b149da69f331ee054/lib/synapse/service_watcher/zookeeper.rb#L235).

This is tricky because we have to be careful to re-set watches if we get disconnected (dealing with the complexity of the pooled zk connections); I remember way back when there used to be a bug where Synapse would lose events, I believe because it lost watches, and I think that one of the ways we fixed it was by ensuring that if we ever lost connection to Zookeeper we immediately bombed out.

We could do what I did for Nerve and even if you bomb out of the watcher you auto-remediate in the main synapse loop and try to relaunch the service_watchers, then we could do something like https://github.com/airbnb/nerve/pull/93 where Synapse could just bomb out of all the dead service_watchers and re-launch them (re-creating the zk connections).

Is your issue just that Synapse is restarting, or that HAProxy is restarting along with it?

lap1817: @jolynch , our issue in this case is Synapse restarting, and it caused some cascading effect so that the restarts repeat for a while.

Actually, my assumption in the earlier reply turns out to be false. 'reconnecting' is NOT only triggered by on_expired_session, but also controlled by the low-level zk client itself. In a test, I triggered packet loss for 20 sec, and saw that the zk client switched from connected to connecting and then back to connected states. So your propose should work in our case. 

So to also include Alexis's comments, we can use

`&& (@zk.associating? || @zk.connected? || @zk.connecting?) `

I believe the associating state is only used when ZK has ACL, and happens as part of the connection establish process. It is good to also include that too.

jolynch: Awesome, I like that test and can believe that (my recollection was that zk-ruby/zookeeper does TCP level reconnects and zk-ruby/zk does session level reconnects and the session state came from the zookeeper but I wasn't sure).

Just curious though, you did _not_ get a session disconnect during that 20s? I vaguely recall something about the underlying zookeeper library having a hard coded session timeout of 10s ([comment](https://github.com/zk-ruby/zk/blob/155db85653c3831cc005140b5fd58adc80a54d8e/lib/zk/client/threaded.rb#L65-L68) for breadcrumb) even if the server is doing 30s but honestly I can't keep track of the ruby's zk client eccentricities there are so many haha.

If you can simulate 20s of packet loss and see the transition but not see the session disconnect I buy that over all else :-)

lap1817: The session expire did happened. 
At first, I emit the state every sec, and was surprised to see that 'expired_session?' is always false even after 30 sec of blocking. Then I added a callback to on_expired_session, and finally saw it. What happened is that,  reconnect was triggered by zk.on_expired_session once the packet blocker is removed. and it happened so quickly that my per-second status output didn't capture

lap1817: @jolynch . I did a further test using actually Synapse with the fix `&& (@zk.associating? || @zk.connected? || @zk.connecting?)` . After blocking the packets for 30 sec, Synapse watcher cleanup is triggered by on_expired_session. If the blocking is shortened to about 10 sec, expired_session didn't happen and cleanup was not triggered.

lap1817:
@jolynch , PTAL

jolynch:
@lap1817 merged, thanks!

PR #251, opened 2018-03-20
Adding `statsd` dependency and exporting some stats

juchem:
Adding support for exporting metrics through `statsd`.
Also, exporting basic metrics about `synapse`'s health and performance.

How it was tested:
unit tests + manual run and checking exported metrics

Reviewers: @lap1817 @alexism

### DIFF
--- a/config/synapse.conf.json
+++ b/config/synapse.conf.json
@@ -1,4 +1,8 @@
 {
+  "statsd": {
+    "host": "localhost",
+    "port": 8125
+  },
   "services": {
     "service1": {
       "default_servers": [
--- a/lib/synapse.rb
+++ b/lib/synapse.rb
@@ -3,6 +3,7 @@ require 'json'
 
 require 'synapse/version'
 require 'synapse/log'
+require 'synapse/statsd'
 require 'synapse/config_generator'
 require 'synapse/service_watcher'
 
@@ -11,8 +12,11 @@ module Synapse
   class Synapse
 
     include Logging
+    include StatsD
 
     def initialize(opts={})
+      StatsD.configure_statsd(opts["statsd"] || {})
+
       # create objects that need to be notified of service changes
       @config_generators = create_config_generators(opts)
       raise "no config generators supplied" if @config_generators.empty?
@@ -29,47 +33,76 @@ module Synapse
       Thread.abort_on_exception = true
 
       log.debug "synapse: completed init"
+    rescue Exception => e
+      statsd && statsd.increment('synapse.stop', tags: ['stop_avenue:abort', 'stop_location:init'])
+      raise e
     end
 
     # start all the watchers and enable haproxy configuration
     def run
       log.info "synapse: starting..."
+      statsd.increment('synapse.start')
 
       # start all the watchers
-      @service_watchers.map { |watcher| watcher.start }
-
-      # main loop
-      loops = 0
-      loop do
-        @service_watchers.each do |w|
-          raise "synapse: service watcher #{w.name} failed ping!" unless w.ping?
-        end
-
-        if @config_updated
-          @config_updated = false
-          @config_generators.each do |config_generator|
-            log.info "synapse: configuring #{config_generator.name}"
-            config_generator.update_config(@service_watchers)
+      statsd.time('synapse.watchers.start.time') do
+        @service_watchers.map do |watcher|
+          begin
+            watcher.start
+            statsd.increment("synapse.watcher.start", tags: ['start_result:success', "watcher_name:#{watcher.name}"])
+          rescue Exception => e
+            statsd.increment("synapse.watcher.start", tags: ['start_result:fail', "watcher_name:#{watcher.name}"])
+            raise e
           end
         end
+      end
 
-        sleep 1
-        @config_generators.each do |config_generator|
-          config_generator.tick(@service_watchers)
+      statsd.time('synapse.main_loop.elapsed_time') do
+        # main loop
+        loops = 0
+        loop do
+          @service_watchers.each do |w|
+            alive = w.ping?
+            statsd.increment('synapse.watcher.ping.count', tags: ["watcher_name:#{w.name}", "ping_result:#{alive ? "success" : "failure"}"])
+            raise "synapse: service watcher #{w.name} failed ping!" unless alive
+          end
+
+          if @config_updated
+            @config_updated = false
+            statsd.increment('synapse.config.update')
+            @config_generators.each do |config_generator|
+              log.info "synapse: configuring #{config_generator.name}"
+              config_generator.update_config(@service_watchers)
+            end
+          end
+
+          sleep 1
+          @config_generators.each do |config_generator|
+            config_generator.tick(@service_watchers)
+          end
+
+          loops += 1
+          log.debug "synapse: still running at #{Time.now}" if (loops % 60) == 0
         end
-
-        loops += 1
-        log.debug "synapse: still running at #{Time.now}" if (loops % 60) == 0
       end
 
     rescue StandardError => e
+      statsd.increment('synapse.stop', tags: ['stop_avenue:abort', 'stop_location:main_loop'])
       log.error "synapse: encountered unexpected exception #{e.inspect} in main thread"
       raise e
     ensure
       log.warn "synapse: exiting; sending stop signal to all watchers"
 
       # stop all the watchers
-      @service_watchers.map(&:stop)
+      @service_watchers.map do |w|
+        begin
+          w.stop
+          statsd.increment("synapse.watcher.stop", tags: ['stop_avenue:clean', 'stop_location:main_loop', "watcher_name:#{w.name}"])
+        rescue Exception => e
+          statsd.increment("synapse.watcher.stop", tags: ['stop_avenue:exception', 'stop_location:main_loop', "watcher_name:#{w.name}"])
+          raise e
+        end
+      end
+      statsd.increment('synapse.stop', tags: ['stop_avenue:clean', 'stop_location:main_loop'])
     end
 
     def reconfigure!
--- a/lib/synapse/config_generator/haproxy.rb
+++ b/lib/synapse/config_generator/haproxy.rb
@@ -1287,7 +1287,7 @@ class Synapse::ConfigGenerator
         return false
       else
         tmp_file_path = "#{opts['config_file_path']}_#{Time.now.to_i}.tmp"
-        File.write(tmp_file_path, data)
+        File.write(tmp_file_path, new_config)
         FileUtils.mv(tmp_file_path, opts['config_file_path'])
         return true
       end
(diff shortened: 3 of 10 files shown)


lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -36,18 +36,22 @@ def start
...
+      statsd.increment('synapse.zk_watcher.clean_stop')
       zk_cleanup
     end
 
     def ping?
       # @zk being nil implies no session *or* a lost session, do not remove
       # the check on @zk being truthy
-      @zk && @zk.connected?
+      alive = @zk && @zk.connected?
+      statsd.increment("synapse.zk_watcher.ping.#{alive ? "success" : "failure"}")
```
alexism: Moving this metric in the main loop would make it available for all watcher types.

lib/synapse.rb
```diff
@@ -29,40 +31,48 @@ def initialize(opts={})
       Thread.abort_on_exception = true
 
       log.debug "synapse: completed init"
+    rescue
+      statsd.increment('synapse.zk_watcher.init.abort')
```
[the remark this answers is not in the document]
juchem: good point

lib/synapse/config_generator/haproxy.rb
```diff
@@ -1287,7 +1287,7 @@ def write_config(new_config)
         return false
       else
         tmp_file_path = "#{opts['config_file_path']}_#{Time.now.to_i}.tmp"
-        File.write(tmp_file_path, data)
+        File.write(tmp_file_path, new_config)
```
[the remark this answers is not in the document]
juchem: good point

config/synapse.conf.json
```diff
@@ -1,4 +1,8 @@
 {
+  "statsd": {
+    "host": "localhost",
+    "port": 9125
```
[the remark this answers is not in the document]
juchem: Yes. I thought it was only true for Datadog, not for `STATSD`, but upon a double check it's the same for both.

lib/synapse/service_watcher/zookeeper.rb
juchem: good point

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -36,18 +36,22 @@ def start
       @zk = nil
 
       log.info "synapse: starting ZK watcher #{@name} @ hosts: #{@zk_hosts}, path: #{@discovery['path']}"
+      statsd.increment('synapse.zk_watcher.started')
       zk_connect
     end
 
     def stop
       log.warn "synapse: zookeeper watcher exiting"
+      statsd.increment('synapse.zk_watcher.clean_stop')
```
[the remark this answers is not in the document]
juchem: yes, `abort`

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -240,37 +248,42 @@ def zk_cleanup
...
-      log.info "synapse: retrieved zk connection to #{@zk_hosts}"
+        # Ensure that all Zookeeper watcher re-use a single zookeeper
+        # connection to any given set of zk hosts.
+        @@zk_pool_lock.synchronize {
+          unless @@zk_pool.has_key?(@zk_hosts)
+            log.info "synapse: creating pooled connection to #{@zk_hosts}"
+            @@zk_pool[@zk_hosts] = ZK.new(@zk_hosts, :timeout => 5, :thread => :per_callback)
+            @@zk_pool_count[@zk_hosts] = 1
+            log.info "synapse: successfully created zk connection to #{@zk_hosts}"
+            statsd.increment('synapse.zk_watcher.connection.created')
```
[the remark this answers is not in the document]
juchem: makes sense

lib/synapse/config_generator/haproxy.rb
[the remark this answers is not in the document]
igor47: looks like this bug dates to https://github.com/airbnb/synapse/pull/248 ! fyi @jolynch and @lap1817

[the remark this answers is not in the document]
juchem: precisely. @lap1817 and I spotted it yesterday

lib/synapse/service_watcher/zookeeper.rb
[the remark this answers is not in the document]
igor47: the dogstatsd way of doing this would be to something like `statsd.increment('synapse.zk_watcher.stop', :tags => ["clean:true"])` and then use `["clean:false"]` in the other case. this lets you more easily slice the metrics later on. in general, it might be helpful to look through this PR with an eye for where using tags would be appropriate

bin/synapse
```diff
@@ -61,6 +61,8 @@ if config.has_key?('service_conf_dir')
   cfiles.each { |x| config['services'][File.basename(x[/(.*)\.(json|yaml)$/, 1])] = parseconfig(x) }
 end
 
+Synapse::StatsD.configure_statsd(config["statsd"] || {})
```
igor47: should probably allow the `Synapse::Synapse::initialize` method do this configuration. the purpose of the bin is to load the `config` object and pass it to `initialize`, but there are other places where you might want to do that (e.g., tests) where you might forget to call `configure_statsd` beforehand

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -113,78 +117,82 @@ def serverset_decode(data)
...
-    def create(path)
+    def create(path, recursion = false)
       log.debug "synapse: creating ZK path: #{path}"
 
-      # recurse if the parent node does not exist
-      create File.dirname(path) unless @zk.exists? File.dirname(path)
-      @zk.create(path, ignore: :node_exists)
+      statsd.time('synapse.zk_watcher.create.time') do
+        # recurse if the parent node does not exist
+        create(File.dirname(path), true) unless @zk.exists? File.dirname(path)
```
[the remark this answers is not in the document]
juchem: I agree. The extra parameter was intended to avoid timing at all steps throughout the recursion, but I forgot to actually act on in. It also complicates the code.

I considered renaming current `create` into an implementation method and creating an entry point that would time that, but I decided to simply time it at the call site. It makes everything simpler and renders this PR less intrusive.

bin/synapse
juchem: Thanks, that's exactly what I wanted. Putting it here was a result of my limited understanding of the code, and I missed this on my cleanup pass. I'll change that.

lib/synapse/statsd.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    class << self
+      include Logging
+
+      @@STATSD_HOST = "localhost"
+      @@STATSD_PORT = 8125
+
+      def statsd_for(classname)
+        log.info "synapse: creating statsd client on host '#{@@STATSD_HOST}' port #{@@STATSD_PORT}"
```
igor47: going to get a lot of identical log lines. maybe include `classname` in the log line? but this could probably be `log.debug`

juchem: I'll do both recommendations.

lib/synapse/service_watcher/zookeeper.rb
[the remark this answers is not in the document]
juchem: I like that. Will change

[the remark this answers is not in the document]
juchem: this seems to also be the wrong place to put this metric.  I'll move it to `synapse.rb -> def run -> ensure`, which represents the proper clean stop place. That will also cover all watchers.

juchem:
PTAL @lap1817

juchem:
cc: @darnaut

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -30,12 +30,17 @@ def initialize(opts={}, synapse)
...
+      # remove domain extents and trailing numbers
+      zk_first_host = zk_host_list.first.split('.').first
+      @zk_cluster = zk_first_host[0..(zk_first_host.rindex(/[^0-9]/) || zk_first_host.length)]
+      @zk_hosts = zk_host_list.join(',')
 
       @watcher = nil
       @zk = nil
 
-      log.info "synapse: starting ZK watcher #{@name} @ hosts: #{@zk_hosts}, path: #{@discovery['path']}"
+      log.info "synapse: starting ZK watcher #{@name} @ cluster #{@zk_cluster} hosts: #{@zk_hosts}, path: #{@discovery['path']}"
```
alexism: The `zk_hosts` string can get pretty long, and the line might get drop depending on the logging facilities (hint: rsyslog).
I think the `zk_cluster` name is a nice addition, and would be sufficient here.

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -30,12 +30,17 @@ def initialize(opts={}, synapse)
     end
 
     def start
-      @zk_hosts = @discovery['hosts'].sort.join(',')
+      zk_host_list = @discovery['hosts'].sort
+      # extract cluster name by filtering name of first host
```
alexism: The host list may contain IP addresses.

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -123,81 +128,89 @@ def create(path)
...
 
-      set_backends(new_backends, new_config_for_generator)
+        set_backends(new_backends, new_config_for_generator)
+      end
     end
 
     # sets up zookeeper callbacks if the data at the discovery path changes
     def watch
       return if @zk.nil?
       log.debug "synapse: setting watch at #{@discovery['path']}"
```
lap1817: can we add a counter metric here so that we know who (`@name`) are watching on a particular service (@discoery['path'])

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -123,81 +128,89 @@ def create(path)
...
-      @zk.children(@discovery['path'], :watch => true).each do |id|
-        begin
-          node = @zk.get("#{@discovery['path']}/#{id}")
-        rescue ZK::Exceptions::NoNode => e
-          # This can happen when the registry unregisters a service node between
-          # the call to @zk.children and @zk.get(path). ZK does not guarantee
-          # a read to ``get`` of a child returned by ``children`` will succeed
-          log.error("synapse: #{@discovery['path']}/#{id} disappeared before it could be read: #{e}")
-          next
+      statsd.time('synapse.watcher.zk.discover.elapsed_time', tags: ["zk_cluster:#{@zk_cluster}", "zk_path:#{@discovery['path']}"]) do
```
lap1817: can we add a counter metric here so that we know who (`@name`) gets update of a particular service (@discoery['path'])

lap1817:
LGTM. with a minor comment of adding one more metric on watchers that are watching a particular service

lib/synapse/service_watcher/zookeeper.rb
juchem: addressed

lib/synapse/service_watcher/zookeeper.rb
juchem: addressed

PR #252, opened 2018-03-22
Add weight support for haproxy config

chase-childers:
Create option for using nerve weights to be backwards compatible, it requires handling the use-case where users implemented HAProxy weighting using the `server_options` or `haproxy_server_options`.  This allows for users to enable the weight functionality with the option of `use_nerve_weights`

Added logic to remove the weight options from `server_options` and `haproxy_server_options` when `use_nerve_weights` is enabled.  This is to prevent producing an odd haproxy backend configuration with multiple weights.

### DIFF
--- a/lib/synapse/config_generator/haproxy.rb
+++ b/lib/synapse/config_generator/haproxy.rb
@@ -814,7 +814,7 @@ class Synapse::ConfigGenerator
       @opts['do_writes'] = true unless @opts.key?('do_writes')
       @opts['do_socket'] = true unless @opts.key?('do_socket')
       @opts['do_reloads'] = true unless @opts.key?('do_reloads')
-
+      @opts['use_nerve_weights'] = true if @opts.key?('use_nerve_weights')
       req_pairs = {
         'do_writes' => 'config_file_path',
         'do_socket' => 'socket_file_path',
@@ -1090,6 +1090,11 @@ class Synapse::ConfigGenerator
             log.info "synapse: restart required because haproxy_server_options changed for #{backend_name}"
             @restart_required = true
           end
+
+          if(@opts['use_nerve_weights'] && old_backend.fetch('weight', "") != backend.fetch('weight', ""))
+            log.info "synapse: restart required because weight changed for #{backend_name}"
+            @restart_required = true
+          end
         end
 
         backends[backend_name] = backend.merge('enabled' => true)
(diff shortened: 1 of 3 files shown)


lap1817:
Can you give provide a bit more context about the change? like who do we need the weight from nerve and how it is used?

lib/synapse/config_generator/haproxy.rb
```diff
@@ -814,7 +814,7 @@ def initialize(opts)
       @opts['do_writes'] = true unless @opts.key?('do_writes')
       @opts['do_socket'] = true unless @opts.key?('do_socket')
       @opts['do_reloads'] = true unless @opts.key?('do_reloads')
-
+      @opts['use_nerve_weights'] = true if @opts.key?('use_nerve_weights')
```
lap1817: what is the expected default behavior when 'use_nerve_weights' is not set in the config? should that be false?

lib/synapse/config_generator/haproxy.rb
```diff
@@ -1153,13 +1158,37 @@ def generate_backend_stanza(watcher, config)
               b = "#{b} cookie #{backend_name}"
             end
           end
-          b = "#{b} #{watcher_config['server_options']}" if watcher_config['server_options'].is_a? String
-          b = "#{b} #{backend['haproxy_server_options']}" if backend['haproxy_server_options'].is_a? String
+
+          if @opts['use_nerve_weights'] && backend['weight'] && ((backend['weight'].is_a? String) || (backend['weight'].is_a? Integer))
```
lap1817: why the 'weight' field can be either a string or a integer? should it just be a integer?

lap1817: or, if string input is really a valid case, should we check whether it represents a valid integer? otherwise, I think ruby will just infer the value as 0 on #1174.

spec/lib/synapse/haproxy_spec.rb
```diff
@@ -208,6 +276,7 @@ class MockWatcher; end;
       expect(haproxy.opts['do_writes']).to eql(true)
       expect(haproxy.opts['do_socket']).to eql(true)
       expect(haproxy.opts['do_reloads']).to eql(true)
+      expect(haproxy.opts['use_nerve_weights']).to eql(nil)
```
lap1817: what is the expected default value?

lib/synapse/config_generator/haproxy.rb
chase-childers: If it is not set in the config, the idea is it should be false.  This is to provide backwards compatibility for this change.  Default to the old behavior.

spec/lib/synapse/haproxy_spec.rb
chase-childers: The expected default value, which is how people would be executing before this change, is to have the flag not exist.

lib/synapse/config_generator/haproxy.rb
chase-childers: I changed this to support only integers.  An invalid string would weight to 0, which could unexpectedly disable a backend.

lib/synapse/config_generator/haproxy.rb
lap1817: ok. so the logic here is to leave it as nil if the key is not set. to make it false by default, shall we do
@opts['use_nerve_weights'] = false unless @opts.key?('use_nerve_weights')

chase-childers: I went ahead and removed this.  This was an artifact of the original change that was not backwards compatible and i swapped this if statement to make the change backwards compatible.  And in the current state, it effectively does nothing, if anything it adds confusion to how the `use_nerve_weights` flag is used.

lap1817:
just to confirm that the last update is on purpose. So you are making the default behavior as: if 'user_nerge_weight' is not set, the expected value is nil

chase-childers:
That is correct.  If `use_nerve_weight` is not set, it will be nil.  All the check's on this argument check its truthy-ness

PR #253, opened 2018-03-28
[Synapse] simply atomic config file output

lap1817:
Since config_generator only runs in the main thread, there is actually no need to timestamp the temporary file. Also avoids the chance to leave stale files in the folder.

Reviewer:
@juchem


juchem:
CI is failing, other than that LGTM

PR #257, opened 2018-04-26
Allow DNS discovery to handle service nodes with an IP address for host

darnaut:
Allow the the `dns` and `zookeeper_dns` service discovery methods to handle service nodes with an IP address for the `host` field.

This facilities migrating services from other discovery methods (such as `zookeeper`) to the `dns` (or `zookeeper_dns`) method by allowing the service discovery method to be changed before publishing new service nodes with hostnames.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    synapse (0.15.3)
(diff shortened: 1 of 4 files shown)


xinyaohu:
@airbnb/traffic @Jason-Jian

PR #258, opened 2018-05-01
Support basic templating in server options

maxburkhardt:
Add support for some basic template interpolation in the `server_options` field. The name, host, and port of a particular backend server are made available for insertion into a `server_options` block, so that a server options configuration can be supplied to synapse that will vary based on the details of a host being configured. This allows options to be set that are dependent on information coming from the service discovery backend.

In particular, this allows one to disable TLS health checks for a TLS backend by specifying `check port <portnumber>`, where `portnumber` varies depending on the backend port that has been discovered by your discovery system.

@lap1817 @juchem @Jason-Jian

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    synapse (0.15.4)
+    synapse (0.16.0)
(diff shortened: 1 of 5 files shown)


lap1817:
LGTM

PR #262, opened 2018-09-28
do not fail for no generator-config-path zk node

jiancheung:
Currently, if `generator_config_path` is pointed to a non-existent zk node, synapse will crash.
Let's fail more gracefully and use the default behavior of `new_config_for_generator = {}`.

@bsherrod @reissbaker 
cc @juchem @lap1817 (common contributors)

### DIFF
--- a/spec/lib/synapse/service_watcher_zookeeper_spec.rb
+++ b/spec/lib/synapse/service_watcher_zookeeper_spec.rb
@@ -133,7 +133,7 @@ describe Synapse::ServiceWatcher::ZookeeperWatcher do
       subject.instance_variable_set('@zk', mock_zk)
       expect(subject.ping?).to be false
     end
-    
+
     context "when generator_config_path is defined" do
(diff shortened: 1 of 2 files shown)


spec/lib/synapse/service_watcher_zookeeper_spec.rb
```diff
@@ -150,6 +150,20 @@
         expect(subject).to receive(:set_backends).with([],{})
         subject.send(:watcher_callback).call
       end
+
+      it 'does not crash if there is no zk node' do
```
melaniecebula: the better way to do this is:
```
context "when there is no zk node" do
  before :each do
    allow(..)
    expect(..)
  end
   ...
   it "does not crash" do
     ...
   end
end
```

This test is also very similar and has a lot of shared `expect` calls with the previous test (it also raises a `ZK::Exceptions::NoNode`) so maybe we can clean that up a bit (ex: put `generator_config_path` in a `let(..)` statement)

PR #270, opened 2018-11-01
Do not crash if file_output not present

Ramyak:
Problem: synapse work if the config is `{"haproxy" => {..}, "file_output"=>{}}`
synapse crashes if config is `{"haproxy" => {..}}`
and does not contain `file_output`

Solution: Test for nil?

Tests: Added test case. Tested on a service.

### DIFF
--- a/lib/synapse/config_generator/file_output.rb
+++ b/lib/synapse/config_generator/file_output.rb
@@ -31,7 +31,7 @@ class Synapse::ConfigGenerator
 
     def update_config(watchers)
       watchers.each do |watcher|
-        next if watcher.config_for_generator[name]['disabled']
+        next if writer_disabled?(watcher)
 
         unless @watcher_revisions[watcher.name] == watcher.revision
           @watcher_revisions[watcher.name] = watcher.revision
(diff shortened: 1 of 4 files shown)


lib/synapse/config_generator/file_output.rb
```diff
@@ -31,6 +31,8 @@ def tick(watchers)
 
     def update_config(watchers)
       watchers.each do |watcher|
+        next if watcher.config_for_generator[name].nil?
```
weeyum: nit:
should we extra this into a helper method?
we can re-use the method for the other change below as well

```
private

def watch_config_disabled?(watcher)
  watcher.config_for_generator[name] && watcher.config_for_generator[name]['disabled']
end
```

lib/synapse/version.rb
```diff
@@ -1,3 +1,3 @@
 module Synapse
-  VERSION = "0.16.1"
+  VERSION = "0.16.2"
```
weeyum: can we bump the gem version in a separate, last commit in this PR? so we can see in logs later what the version contains.

PR #279, opened 2019-03-12
warn when parsing a unknown generator name

austin-zhu:
When parsing an unknown `generator` from service configuration stored on `zookeeper`, `synapse` skips it but still throw `error`. `zookeeper` is designed as a generic datastore for service discovery data but not only for synapse. Therefore, there might be any random key-value pair stored under `zookeeper` node, instead of those known by synapse. This is expected and reasonable. Instead of logging as `error` level, it should make more sense to just be a `warn`.

Also, I have to limit `travis` script to use `bundler 1.17.3` in order to support legacy ruby version.
```
ERROR:  Error installing bundler:
	The last version of bundler (>= 0) to support your Ruby & RubyGems was 1.17.3. Try installing it with `gem install bundler -v 1.17.3`
	bundler requires Ruby version >= 2.3.0. The current ruby version is 1.9.1.
```

@Jason-Jian @allenlsy @chase-childers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ rvm:
   - 2.3.1
 
 before_install:
-  - gem install bundler
+  - gem install bundler -v 1.17.3
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    synapse (0.16.2)
+    synapse (0.16.3)
       aws-sdk (~> 1.39)
       docker-api (~> 1.7)
       dogstatsd-ruby (~> 3.3.0)
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -334,7 +334,7 @@ class Synapse::ServiceWatcher
       #   each value should be a hash (could be empty)
       decoded.collect.each do |generator_name, generator_config|
         if !@synapse.available_generators.keys.include?(generator_name)
-          log.error "synapse: invalid generator name in ZK node at #{@discovery['path']}:" \
+          log.warn "synapse: invalid generator name in ZK node at #{@discovery['path']}:" \
             " #{generator_name}"
           next
         else
--- a/lib/synapse/version.rb
+++ b/lib/synapse/version.rb
@@ -1,3 +1,3 @@
 module Synapse
-  VERSION = "0.16.2"
+  VERSION = "0.16.3"
 end


jolynch:
So I'm not that familiar with how Airbnb is using the dynamic zookeeper watcher configuration feature (last I was at Yelp we didn't use it since we autogenerated the config.json dynamically anyways; Feng Pan et al have way more context then me though), but if I understand it properly the configuration for each watcher is now stored in a znode which are looked up dynamically at a certain path (looks like the `generator_config_path` configuration element). Based on that I would assume that the user has full control on e.g. namespacing that path and ensuring that it's not a shared path with other tenants on the cluster.

Given that the user controls the input, are there valid reasons for them to include configuration for config generators that don't exist? What's an example of such a zk layout?

Personally I may actually prefer Synapse _crash_ if given invalid configuration as opposed to logging either an error or a warning (as I believe Synapse would crash if given invalid `config_generators` in `config.json`). This of course is predicated on invalid configuration being considered a bug and assuming that Synapse may crash at any time.

austin-zhu:
@jolynch thanks for quick looking into. We are storing HAProxy config in our zookeeper node for some cases. But there will be other usage soon that we are going to store other `non` HAProxy config to zk_node. To reduce noise of Synapse log, I have foreseen this issue and adjust it to `warn`.

In other words, zookeeper is just the data store used by synapse. It could definitely get a chance to store some other service discovery data which is not for synapse, as long as the zk cluster is not sharded to serve synapse only. Therefore, instead of crashing synapse and stop parsing the config, it should be better to just gracefully skip the unknown generator name. And I think this is not for `airbnb` only, but also apply for other scenario.

austin-zhu:
@jolynch since Jason has approved, I will wait for your opinion for a while before merging.

jolynch:
@austin-zhu Please don't wait for me, I assume you're making this change because you need it so shipit to prod :+1: 
Also I realize it really would have been better for me to have commented when the original feature logged instead of crashing rather than now, just noticed it an was curious.

>In other words, zookeeper is just the data store used by synapse. It could definitely get a chance to store some other service discovery data which is not for synapse, as long as the zk cluster is not sharded to serve synapse only. Therefore, instead of crashing synapse and stop parsing the config, it should be better to just gracefully skip the unknown generator name. I do agree with you that this will indicate the path is not for smartstack only but shared with other tenants who will write data to the cluster.

I see what you're saying. I think the context I was missing is I assumed these config znodes were namespaced, e.g. [`generator_config_path`](https://github.com/airbnb/synapse/blob/dc21a3c19acbac55268ded507153e8f5f821c2bc/lib/synapse/service_watcher/zookeeper.rb#L195) might be set to `/smartstack/haproxy_cfg.json` or `/smartstack/file_and_haproxy_cgf.json` so you could easily put the non smartstack state you want into `/smartstack/extra_metadata.json`. It sounds, however, that you've got `generator_config_path` all set to the same znode like `/smartstack/generator_config.json` and now you want to change the data that's in `/smartstack/generator_config.json`? I am still slightly confused why you wouldn't just add `/smartstack/extra_metadata.json` or something but maybe you've got some automation (Chef?) or something which knows how to write that one znode but modifying it to write to two znodes may be harder?

austin-zhu:
@jolynch I will merge it, thanks.

In Airbnb, we are putting `generator_config_path` as `nil` so that synapse loads data directly from its zk discovery path. The way we put data is to the `data` field under a node like `/production/services/foo/services`. Sorry I could tell you more what else than Synapse data we stored there. We did think about write to a different znode or even different zk cluster but it will our `consumer` `read` logic much more complex. Therefore, we decide to store them in a single centralized place, together with synapse data.

jolynch:
Ah I see, thanks for sharing the context :)

PR #280, opened 2019-04-09
Add failure counters for updating config

anson627:
config_generator.update_config has several steps are subjected to failure:
1. read existing config file
2. generate new config
3. write temp config file
4. replace existing config file

we need catch exception on this code path, emit success & failure counter, and log exception details

@austin-zhu @Jason-Jian @ken-experiment


anson627:
verified error log and metrics by simulating file I/O exception

PR #282, opened 2019-04-11
Add failure counters for zookeeper watcher

anson627:
Add failure counters metrics for zookeeper:
1. register watcher failure
2. get data failure
3. deserialize data failure

@austin-zhu @ken-experiment @Jason-Jian


lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -200,6 +205,7 @@ def discover
             new_config_for_generator = {}
           rescue StandardError => e
             log.error "synapse: invalid config data in ZK node at #{@discovery[discovery_key]}: #{e}"
```
austin-zhu: add `skip XXX node`

PR #284, opened 2019-04-16
Add retry logics for zookeeper get

anson627:
zookeeper get can return nil or throw timeout exception, when network is flaky or zookeeper is overloaded.

adding retry logics (by default 3 times with 5 second interval) can alleviate the problem.

if all retries failed, throw runtime error forcing the process to re-establish zookeeper connections.

@austin-zhu @ken-experiment

### DIFF
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -135,22 +135,30 @@ class Synapse::ServiceWatcher
     end
 
     # get with retry to reduce failure when network connectivity is flaky or ZK is overloaded
-    def get(path, opts = {})
-      retry_limit = opts[:retry_limit] || 3
-      opts.delete(:retry_limit)
-      retry_interval = opts[:retry_interval] || 5
-      opts.delete(:retry_interval)
+    def zk_get_path(path, opts = {})
+      retry_limit = 3
+      if opts.has_key? :retry_limit
+        retry_limit = opts[:retry_limit]
+        opts.delete(:retry_limit)
+      end
+
+      retry_interval = 5
+      if opts.has_key? :retry_interval
+        retry_interval = opts[:retry_interval]
+        opts.delete(:retry_interval)
+      end
+
       retry_count = 0
       begin
-        if retry_count > retry_limit
-          raise RuntimeError, "synapse: get ZK path failed after retry"
-        end
-        retry_count += 1
         node = @zk.get(path, opts)
         raise IOError, "synapse: get ZK path return nil" if node.nil?
         return node
       rescue ZK::Exceptions::OperationTimeOut, IOError => e
         log.error("synapse: get ZK #{path} failed for the #{retry_count} time: #{e}")
+        retry_count += 1
+        if retry_count > retry_limit
+          raise RuntimeError, "synapse: get ZK path failed after retry"
+        end
         sleep retry_interval
         retry
       end
@@ -166,7 +174,7 @@ class Synapse::ServiceWatcher
         @zk.children(@discovery['path'], :watch => true).each do |id|
           begin
             node = statsd_time('synapse.watcher.zk.get.elapsed_time', ["zk_cluster:#{@zk_cluster}", "service_name:#{@name}"]) do
-              get("#{@discovery['path']}/#{id}")
+              zk_get_path("#{@discovery['path']}/#{id}")
             end
           rescue ZK::Exceptions::NoNode => e
             # This can happen when the registry unregisters a service node between
@@ -219,7 +227,7 @@ class Synapse::ServiceWatcher
         if discovery_key
           begin
             node = statsd_time('synapse.watcher.zk.get.elapsed_time', ["zk_cluster:#{@zk_cluster}", "service_name:#{@name}"]) do
-              get(@discovery[discovery_key], :watch => true)
+              zk_get_path(@discovery[discovery_key], :watch => true)
             end
             new_config_for_generator = parse_service_config(node.first)
           rescue ZK::Exceptions::NoNode => e
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -134,6 +134,28 @@ class Synapse::ServiceWatcher
       @zk.create(path, ignore: :node_exists)
     end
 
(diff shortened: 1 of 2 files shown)


lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -134,6 +134,28 @@ def create(path)
...
+      retry_limit = opts[:retry_limit] || 3
+      opts.delete(:retry_limit)
+      retry_interval = opts[:retry_interval] || 5
+      opts.delete(:retry_interval)
+      retry_count = 0
+      begin
+        if retry_count > retry_limit
+          raise RuntimeError, "synapse: get ZK path failed after retry"
+        end
+        retry_count += 1
```
austin-zhu: this should happen before `if retry_count > retry_limit`? actually, you will end up trying `4` times in total now. Or maybe you actually mean try once first then retrying another 3 times? So to make logic simple, you could move retry check into `rescue` part. Following example is trying up to `4` times.

```
retry_count += 1
if retry_count > retry_limit
   raise RuntimeError, "synapse: get ZK path failed after retry"
else
  sleep retry_interval
  retry
end
```

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -134,6 +134,28 @@ def create(path)
       @zk.create(path, ignore: :node_exists)
     end
 
+    # get with retry to reduce failure when network connectivity is flaky or ZK is overloaded
+    def get(path, opts = {})
```
austin-zhu: could we rename this function to `zk_get_path` or something more concrete?

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -134,6 +134,28 @@ def create(path)
...
+      retry_count = 0
+      begin
+        if retry_count > retry_limit
+          raise RuntimeError, "synapse: get ZK path failed after retry"
+        end
+        retry_count += 1
+        node = @zk.get(path, opts)
+        raise IOError, "synapse: get ZK path return nil" if node.nil?
+        return node
+      rescue ZK::Exceptions::OperationTimeOut, IOError => e
```
austin-zhu: so we only `rescue` this two types of exception? or simply retry on all possible zk error? However, we should only retry on `zk` errors.

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -134,6 +134,28 @@ def create(path)
       @zk.create(path, ignore: :node_exists)
     end
 
+    # get with retry to reduce failure when network connectivity is flaky or ZK is overloaded
+    def get(path, opts = {})
+      retry_limit = opts[:retry_limit] || 3
+      opts.delete(:retry_limit)
+      retry_interval = opts[:retry_interval] || 5
```
austin-zhu: `5` sec might be so long. We could try `1` `2` `3` respectively during three tries.

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -134,6 +134,28 @@ def create(path)
       @zk.create(path, ignore: :node_exists)
     end
 
+    # get with retry to reduce failure when network connectivity is flaky or ZK is overloaded
+    def get(path, opts = {})
+      retry_limit = opts[:retry_limit] || 3
+      opts.delete(:retry_limit)
```
austin-zhu: NIT: though `opts[:retry_limit] is nil` won't throw exception when `delete`, it might not worth to do this check two times. You could simply do

```
retry_limit = 3
if opts[:retry_limit]
  retry_limit = opts[:retry_limit]
  opts.delete(:retry_limit)
end
```

austin-zhu: hmm, actually, why do we need to delete `:retry_limit`? it will be still used for further fetch zk path, right?

austin-zhu:
also, remember to bump version

lib/synapse/service_watcher/zookeeper.rb
anson627: sure, thanks for the suggestion!

lib/synapse/service_watcher/zookeeper.rb
anson627: these two are obviously retriable error, I was able to reproduce by cutting off the network. According to ZK API documentation [here](https://www.rubydoc.info/gems/zk/ZK/Client/Base#get-instance_method), it could also throw ZK::Exceptions::NoNode, but this means node does not exist any more, and we should instead do some cleanup. So here we just throw the error, and the call already has cleanup logics.

anson627: the common practice is to differentiate retriable vs non-retriable errors, handle them separately.

lib/synapse/service_watcher/zookeeper.rb
anson627: To find reasonable retry interval, we have to experiment out. There is some trade-off. If retry interval is too small, we might make things worse by adding more load when zookeeper is already overloaded. If retry interval is too big, it obviously delay the time to recover from intermittent network connectivity issue.

You are right, exponential back-off is probably a more elegant solution. Before throwing sophisticated solution here, I want to try out something simple first.

lib/synapse/service_watcher/zookeeper.rb
anson627: sounds good!

anson627: Here I hesitate to pass some parameters to zookeeper client API while zookeeper doesn't understand what it means. I only want to pass along the rest of parameters specified by caller. And I'm new to Ruby, and couldn't find a better way to express that after some research. If you know, please let me know.

lib/synapse/service_watcher/zookeeper.rb
anson627: retry_limit means how many times it should retry on top of the initial try. Let me document the input to make it obvious.

> also, remember to bump version

Sure

anson627:
> This change LGTM.
> 
> Just want to get more context, when throw exceptions instead of swallowing them in the previous code, will it introduce any unexpected behaviors in caller functions?

The caller's behaviors stay the same. After retries several times specified by caller, it will still throw error.

However errors are handled by two callers inconsistently, based on original implementation, one would raise, the other would swallow. For that, I hesitate to change it yet, because I don't think I fully understand the intention.

PR #287, opened 2019-05-02
Fix synapse watch lost during zookeeper server reboot

anson627:
we recently discover synapse 'stuck' issue in production: during zookeeper server reboot, watcher was lost forever for one zookeeper path

The key takeways are here:
1. zookeeper watcher is one-time trigger, if one watcher is lost and not re-set, any future events associated will not be notified.
2. zookeeper watcher setting involves two steps: 1. register watcher if it doesn't exist, 2 enable watcher by get/children call (with :watch = true)
2. during zookeeper server restart, client can potentially enter two states: session expire or reconnecting, the currently implementation doesn't handle reconnecting.
3. watcher can get lost when client is disconnected from server.

The fix is to listen on reconnect state, and then re-register and re-enable watcher.

Also log entry when watchers are registered and enabled (both parent and children), for better observability in future operation

@austin-zhu @Jason-Jian

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    synapse (0.16.5)
+    synapse (0.16.6)
       aws-sdk (~> 1.39)
       docker-api (~> 1.7)
       dogstatsd-ruby (~> 3.3.0)
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -254,7 +254,10 @@ class Synapse::ServiceWatcher
       log.debug "synapse: setting watch at #{@discovery['path']}"
 
       statsd_time('synapse.watcher.zk.watch.elapsed_time', ["zk_cluster:#{@zk_cluster}", "zk_path:#{@discovery['path']}", "service_name:#{@name}"]) do
-        @watcher = @zk.register(@discovery['path'], &watcher_callback) unless @watcher
+        unless @watcher
+          @watcher = @zk.register(@discovery['path'], &watcher_callback)
+          log.info "synapse: register watcher at path #{@discovery['path']}"
+        end
 
         # Verify that we actually set up the watcher.
         unless @zk.exists?(@discovery['path'], :watch => true)
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -256,7 +256,7 @@ class Synapse::ServiceWatcher
       statsd_time('synapse.watcher.zk.watch.elapsed_time', ["zk_cluster:#{@zk_cluster}", "zk_path:#{@discovery['path']}", "service_name:#{@name}"]) do
         unless @watcher
           @watcher = @zk.register(@discovery['path'], &watcher_callback)
-          log.info "synapse: register watcher at path #{@discovery['path']}"
+          log.info "synapse: register watch at #{@discovery['path']}"
         end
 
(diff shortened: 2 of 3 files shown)


anson627:
For more details here:

TL;DR;
The same session will be reestablished when the ZK client reconnects to the ZK cluster/server. Once it happens, there is no guarantee on the server side to keep any watcher object even if the client side still have the corresponded watcher.
What happened
During ZK operation (ZK host replacement), we need to restart every single ZK host in order to synchronize all ZK hosts and make them aware of the change. This will lead to all current connections interrupted.

When a client (session) becomes partitioned from the ZK cluster, it will begin searching the list of servers that were specified during session initialization, a.k.a., first time creation. Eventually, when connectivity between the client and at least one of the servers is re-established, the same session will either again transition to the connected state (if reconnected within the session timeout value) or it will transition to the expired state (if reconnected after the session timeout).

From Zookeeper Programming Guide: In general, when a client reconnects, any previously registered watches will be reregistered and triggered if needed. However, here is one case where a watch may be missed -- a watch for the existence of a znode not yet created will be missed if the znode is created and deleted while disconnected.

At least, even from the official doc, there is a case a watcher might be missed by ZK server. However, in our Synapse implementation, we won’t recreate the missed watcher only until session expired. In other words, if a connection could be reestablished within timeout (5 sec), we couldn’t handle watcher missing case until Synapse crashes for some other reason later. Before it, the client side watcher enters into a zombie mode and won’t get any update from ZK server, due to the watcher disappearing from the server.
Possible solutions
Proactively prevent missing watcher
Other than on_expired_session callback, ZK client also provides on_connecting and on_connected callbacks. Either of them will give Synapse a way to proactively catch reconnection appearing, so that Synapse could recklessly register all watchers again to avoid server side missing case.
References
https://github.com/python-zk/kazoo/issues/151
https://bowenli86.github.io/2016/09/15/distributed%20system/zookeeper/ZooKeeper-Sessions-and-Session-Management/
https://zookeeper.apache.org/doc/r3.3.3/zookeeperProgrammers.html

anson627:
Tested by restarting zookeeper server, and see reconnecting events:

I, [2019-05-06T17:46:29.596306 #6539]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: ZK client is attempting to reconnect mango-production
I, [2019-05-06T17:46:29.596392 #6539]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: ZK client is attempting to reconnect mango-production
I, [2019-05-06T17:46:29.596417 #6539]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: ZK client is attempting to reconnect mango-production
I, [2019-05-06T17:46:36.280941 #6539]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: ZK client has reconnected mango-production
I, [2019-05-06T17:46:36.323345 #6539]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: set watch for parent at /production/services/mango-production/services
I, [2019-05-06T17:46:36.323501 #6539]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: discovering backends for service mango-production
I, [2019-05-06T17:46:36.339719 #6539]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: set watch for children at /production/services/mango-production/services
I, [2019-05-06T17:46:36.339818 #6539]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: backends for service mango-production do not change.

@austin-zhu

PR #288, opened 2019-05-22
Remove on_connecting for less threads

anson627:
due to zk-client implementation, it keeps a single thread open after registering the callback in order to guarantee high concurrency https://github.com/zk-ruby/zk/wiki/EventDeliveryModel

we need keep number of callback per watch as few as possible.

@austin-zhu @Jason-Jian

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    synapse (0.16.6)
+    synapse (0.16.7)
       aws-sdk (~> 1.39)
       docker-api (~> 1.7)
       dogstatsd-ruby (~> 3.3.0)
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -337,12 +337,6 @@ class Synapse::ServiceWatcher
           zk_cleanup
         end
 
-        # handle session reconnecting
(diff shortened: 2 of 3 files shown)


anson627:
> Also bump the version?

Done

igor47:
hmm i would've liked to see more justification for this change. how often was this callback triggered? what resource are you saving, and how impactful is this change?

anson627:
A thread with an event queue is allocated if a callback is registered, no matter if callback is triggered or not. The number of threads increase linearly as number of watchers. If there are 100 watchers, that adds 100 threads.

igor47:
so the overhead is in memory? what's the memory overhead of one thread? what do typical savings look like after removing this callback?

was this callback not useful? did you ever see it triggered? how do you know, going forward, if the connection to ZK dropped and was re-established?

anson627:
A constantly running thread has both memory and CPU overhead. When you have thousands of watchers and thousands more threads, the context switching and garbage collection become noticeable.

the on_connecting callback was useful for debugging, when ZK server restarts, we can see clients reconnecting (on_connecting gets triggered) and get reconnected (on_connected) within seconds. We keep on_connected callback to reset watcher when the connection is re-established and ready for communication:
https://github.com/airbnb/synapse/pull/288/files#diff-86201d0784bc64799b300dce40bd1d45L348

PR #292, opened 2019-06-13
Improve on_connected callback to avoid reconnect storm

anson627:
The on_connected callback was originally introduced to handle an issue that zookeeper session got stuck (will not receive zookeeper watch event forever), during a connection lost and reconnect.

But we discovered some side-effect of this fix, there are multiple on_connected callback get triggered for a single disconnect and reconnect (here seems other people are facing the same issue: https://github.com/zk-ruby/zk-eventmachine/issues/2). And since we do reset zookeeper watcher by read everything. This creates substantial load on zookeeper.

Here are two optimizations introduced by this change:
1. ensure there is only one watcher refresh within a time window (e.g. 60 seconds) no matter how many on_connected callback gets triggered
2. optimize the watcher refresh by just re-enable (parent and children) watcher, to avoid zk get storm

@austin-zhu @Jason-Jian @allenlsy @Ramyak 

Before (reboot zookeeper twice, got 3 on_connected each time):
image

After (reboot zookeeper server once, got watcher refresh once):
Screen Shot 2019-06-13 at 7 39 21 PM

reboot zookeeper server twice, only got watcher refresh twice
Screen Shot 2019-06-13 at 8 48 56 PM

### DIFF
--- a/lib/synapse/service_watcher/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper.rb
@@ -367,12 +367,14 @@ class Synapse::ServiceWatcher
         # http://zookeeper.apache.org/doc/r3.3.5/zookeeperProgrammers.html#ch_zkSessions
         @zk.on_connected do
           log.info "synapse: ZK client has reconnected #{@name}"
-          # random backoff to avoid checking and refreshing all watchers at the same time
-          sleep rand(10)
           unless test_and_set_reconnect_time(Time.now)
             log.info "synapse: ZK client skip since last reconnect is too close #{@name}"
             return
           end
+
+          # random backoff to avoid checking and refreshing all watchers at the same time
+          sleep rand(30)
+
           # zookeeper watcher is one-time trigger, and can be lost when disconnected
           # https://zookeeper.apache.org/doc/r3.3.5/zookeeperProgrammers.html#ch_zkWatches
           # only need re-enable watcher on parent path and children list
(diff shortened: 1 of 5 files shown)


lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -363,13 +363,24 @@ def zk_connect
...
+          end
+          # test-and-set should be thread safe based on per-callback model
+          # https://github.com/zk-ruby/zk/wiki/EventDeliveryModel
+          @last_reconnect_time = now
           # zookeeper watcher is one-time trigger, and can be lost when disconnected
           # https://zookeeper.apache.org/doc/r3.3.5/zookeeperProgrammers.html#ch_zkWatches
-          @watcher.unsubscribe unless @watcher.nil?
-          @watcher = nil
-          watcher_callback.call
+          # only need reanble watcher on parent path and children list
```
allenlsy: typo on re-enable

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -363,13 +363,24 @@ def zk_connect
         # http://zookeeper.apache.org/doc/r3.3.5/zookeeperProgrammers.html#ch_zkSessions
         @zk.on_connected do
           log.info "synapse: ZK client has reconnected #{@name}"
-          # random backoff to avoid refresh at the same time
+          # random backoff to avoid checking and refreshing all watchers at the same time
           sleep rand(10)
+          now = Time.now
+          # ensure there is only one refresh can happen within a time window
+          if !@last_reconnect_time.nil? && (now - @last_reconnect_time) < 60
```
Ramyak: Can we add a test case for this if else logic?

PR #299, opened 2019-09-25
Add retry policy to zookeeper network calls

anson627:
# Problem
When zookeeper call fails (due to flaky network or server overload), synapse crashes immediately, restarts and reads everything from everywhere. This causes `retry storm` and `cascading failure`: zookeeper server struggles to recover from `failure` under heavy load.

# Analysis
Here is list of top exceptions
Screen Shot 2019-09-23 at 3 36 36 PM

# Solution
When client tries to establish zookeeper connection or make zookeeper calls, it should retry with exponential back-off upon retriable errors (e.g. connection loss, operation timeout)

# Test
local smoke test retry until success
```
I, [2019-09-25T13:05:10.919287 #4167]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: discovering backends for service mango-test
D, [2019-09-25T13:05:30.924689 #4167] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk list children at /mango-test/services for 1 times
D, [2019-09-25T13:06:03.932508 #4167] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk list children at /mango-test/services for 2 times
D, [2019-09-25T13:06:36.935165 #4167] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk list children at /mango-test/services for 3 times
D, [2019-09-25T13:07:09.942464 #4167] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk list children at /mango-test/services for 4 times
D, [2019-09-25T13:07:42.950754 #4167] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk list children at /mango-test/services for 5 times
D, [2019-09-25T13:08:15.957269 #4167] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk list children at /mango-test/services for 6 times
I, [2019-09-25T13:08:29.323375 #4167]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: backends for service mango-test do not change.
```
local smoke test retry until failure
```
D, [2019-09-25T13:09:51.029783 #4839] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk create at /mango-test/services for 1 times
D, [2019-09-25T13:10:24.040245 #4839] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk create at /mango-test/services for 2 times
D, [2019-09-25T13:10:57.044755 #4839] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk create at /mango-test/services for 3 times
D, [2019-09-25T13:11:30.051532 #4839] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk create at /mango-test/services for 4 times
D, [2019-09-25T13:12:03.053940 #4839] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk create at /mango-test/services for 5 times
D, [2019-09-25T13:12:36.059183 #4839] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk create at /mango-test/services for 6 times
D, [2019-09-25T13:13:09.070445 #4839] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk create at /mango-test/services for 7 times
D, [2019-09-25T13:13:42.075808 #4839] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk create at /mango-test/services for 8 times
D, [2019-09-25T13:15:12.080919 #4839] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk create at /mango-test/services for 9 times
D, [2019-09-25T13:16:42.086014 #4839] DEBUG -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zk create at /mango-test/services for 10 times
E, [2019-09-25T13:17:12.091733 #4839] ERROR -- Synapse::Synapse: synapse: encountered unexpected exception #<Zookeeper::Exceptions::ContinuationTimeoutError: response for meth: :exists, args: [9, "/mango-test", nil, nil], not received within 30 seconds> in main thread
W, [2019-09-25T13:17:12.091869 #4839]  WARN -- Synapse::Synapse: synapse: exiting; sending stop signal to all watchers
W, [2019-09-25T13:17:12.091911 #4839]  WARN -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zookeeper watcher exiting
I, [2019-09-25T13:17:12.091942 #4839]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zookeeper watcher cleaning up
I, [2019-09-25T13:17:12.091981 #4839]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: closing zk connection to 127.0.0.1:2181
I, [2019-09-25T13:17:12.093436 #4839]  INFO -- Synapse::ServiceWatcher::ZookeeperWatcher: synapse: zookeeper watcher cleaned up successfully
/Users/anson/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/zookeeper-1.4.11/lib/zookeeper/continuation.rb:117:in `block in value': response for meth: :exists, args: [9, "/mango-test", nil, nil], not received within 30 seconds (Zookeeper::Exceptions::ContinuationTimeoutError)
```
(TODO) stress testing in zookeeper test environment

#Reviewers
@Ramyak @Jason-Jian @austin-zhu @panchr


lib/synapse/retry_policy.rb
```diff
@@ -0,0 +1,37 @@
...
+      rescue *retriable_errors => error
+        if attempts >= max_attempts
+          raise error
+        end
+        sleep get_retry_interval(base_interval, max_interval, attempts, Time.now - start_time)
+        retry
+      end
+    end
+
+    def get_retry_interval(base_interval, max_interval, attempts, elapsed)
```
panchr: What is the point in subtracting out `elapsed` here?

It adds in additional complexity (in terms of reasoning about how long the `n`th attempt will wait for). In addition, I don't think that's standard in typical exponential back-off algorithms.

spec/lib/synapse/retry_policy_spec.rb
```diff
@@ -0,0 +1,150 @@
+require 'spec_helper'
+require 'synapse/retry_policy'
+
+describe "test with retry" do
+  include Synapse::RetryPolicy
+
+  it "default retry succeed" do
+    expected_attemptes = 1
```
panchr: nit: extra `e` in `expected_attempts` (this occurs in a few places below as well)

spec/lib/synapse/service_watcher_zookeeper_spec.rb
```diff
@@ -155,13 +117,103 @@
...
       expect(mock_zk).to receive(:get).with('some/path', {:watch=>true}).and_return("")
-      expect(mock_zk).to receive(:get).with('some/path/test_child_1', {}).and_raise(ZK::Exceptions::NoNode)
+      expect(mock_zk).to receive(:get).with('some/path/test_child_1').and_raise(ZK::Exceptions::NoNode)
 
       subject.instance_variable_set('@zk', mock_zk)
       expect(subject).to receive(:set_backends).with([],{})
       subject.send(:watcher_callback).call
     end
 
+    context 'retry watcher_callback' do
```
panchr: In Rspec, you use `describe` to label *what* you are testing, `context` to label *when* that test applies (i.e. the conditions that it checks) and `it` to label the *expectation*. 

In this case, you want to use `describe watcher_callback`, `context 'when it fails'`, `it 'retries until success'` or something like that. While that doesn't change the behavior of the tests, it does make the tests easier to read :)

spec/lib/synapse/service_watcher_zookeeper_spec.rb
```diff
@@ -155,13 +117,103 @@
...
+        expect(mock_zk).to receive(:register)
+        expect(mock_zk).to receive(:exists?).with('some/path', {:watch=>true}).once.and_return(true)
+        expect(mock_zk).to receive(:children).with('some/path', {:watch=>true}).once.and_return(["test_child_1"])
+        expect(mock_zk).to receive(:get).with('some/path/test_child_1').exactly(2).and_raise(::Zookeeper::Exceptions::NotConnected)
+        subject.instance_variable_set('@zk', mock_zk)
+        expect { subject.send(:watcher_callback).call }.to raise_error(::Zookeeper::Exceptions::NotConnected)
+      end
+    end
+
+    context 'retry start' do
```
panchr: same as above

lib/synapse/retry_policy.rb
anson627: If we don't consider how long it takes a call to fail, the interval btw retries can be skewed. Say we want to delay 60 seconds btw retires. A call failed with connection lost can only take 1 second, the retry interval becomes 61 seconds. A call failed with operation timeout can take 30 seconds, the retry interval becomes 90 seconds. As retry interval accumulates, the total delay of API call can be skewed significantly.

Agree this is not standard. But it is quite common variant to take actual call delay into consideration:
https://github.com/cenkalti/backoff/blob/2146c93394225c3732078705043ce9f26584d334/exponential.go#L128

https://github.com/apache/incubator-gobblin/blob/master/gobblin-utility/src/main/java/org/apache/gobblin/util/ExponentialBackoff.java#L72

As we care about the upper bound delay of retry, we can expose another parameter called `max_delay` to stop retry once total elapsed reached the limit.

spec/lib/synapse/service_watcher_zookeeper_spec.rb
anson627: right now most of existing tests use `describe` to label which class to test, `context` to label which `method`, and `it` to label both condition and expectation.

agree this is not ideal, but at least I can make it consistent with existing tests.

panchr: Okay, that's fair

lib/synapse/retry_policy.rb
```diff
@@ -0,0 +1,45 @@
+module Synapse
+  module RetryPolicy
+    def with_retry(options = {}, &callback)
+      max_attempts = options['max_attempts'] || 1
+      max_delay = options['max_delay'] || Float::INFINITY
```
austin-zhu: if not set, we are going to have infinite delay? which means we will hang on here?

lib/synapse/retry_policy.rb
```diff
@@ -0,0 +1,45 @@
...
+      base_interval = options['base_interval'] || 0
```
austin-zhu: what about `first_interval`?

austin-zhu: what is the unit? I assume it is `millisecond`? might put some notes in the readme.

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -35,6 +42,12 @@ def initialize(opts={}, synapse)
           @decode_method = self.method(:serverset_decode)
         end
       end
+
+      @retry_policy = @discovery['retry_policy'] || {}
```
austin-zhu: just want to make sure that you want to exponential backoff on by default.

anson627: no retry by default. if `retry_policy` is not in the config, @retry_policy will be empty, `with_retry` by default set `max_attempts` to ` 1, so only original call will be attempted.

lib/synapse/retry_policy.rb
anson627: it is seconds, let me document in the readme to make it obvious.

lib/synapse/retry_policy.rb
anson627: change to 3600 seconds

PR #304, opened 2019-11-08
Emit version with ping metric

panchr:
This adds metrics about the Synapse version per-ping. Tested by checking that metrics are properly emitted on a specific host.


anson627:
LGTM

PR #309, opened 2020-02-24
[WIP] Add MultiWatcher for watching multiple sources

panchr:
## Summary
Add in a new `MultiWatcher` watcher type which can create multiple "child" watchers and then resolve their entries.

## PRs
- [x] #310
- [ ] #311 
- [x] #312 
- ~[ ] #313~
- [x] #314 
- [x] #315 

## Todo
- [ ] reducer method for `ping?` and backends
- [ ] configuration for reducer method
- [ ] documentation

## Tests
- [ ] unit tests
- [ ] creating multiple Zookeeper watchers
- [ ] resolution between multiple Zookeeper watchers

## Reviewers
TODO:
```
@austin-zhu @lcharignon @bsherrod 
```

### DIFF
--- a/config/discovery-multi.conf.json
+++ b/config/discovery-multi.conf.json
@@ -1,0 +1,93 @@
+{
+  "services": {
+    "service1": {
+      "discovery": {
+	"method": "zookeeper",
+	"path": "/services/service1",
+	"discovery_jitter": 25,
+	"hosts": [
+	  "localhost:2181",
+	  "localhost:2182",
+	  "localhost:2183"
+	]
+      },
+      "discovery_multi": {
+	"resolver": {
+	  "method": "fallback"
+	},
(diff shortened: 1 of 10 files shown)


panchr:
Fleshing out the idea a bit more: the `reduce` function will accept the children watchers and decide how to get the backends from them.

For example, to always choose the `primary` ones:
```ruby
def reduce(watchers)
    return watchers['primary'].backends
end
```

We may also want to implement `ping?` to do something similar: have a way to `reduce` the different `ping?` results into one.

I'm not sure what the configuration for this would look like: perhaps we can have multiple `Reducer` classes (similar to the multiple `Watcher` classes) which each implement the API. That would the cleanest, in my opinion.

panchr:
Tracked under #311

PR #310, opened 2020-02-28
Allow for configurable notification callback for watchers

panchr:
## Summary
This PR allows for configuring a custom notification callback when a watcher has an update, instead of always called `synapse.reconfigure!` by default.

Note that the `BaseWatcher.reconfigure!` method essentially acts as a notification of an update: this triggers the main Synapse object to update the config generators (i.e. create new HAProxy configuration and reload the process).

This will be used as part of the multi-watcher (WIP: https://github.com/airbnb/synapse/pull/309), in order to allow the multi-watcher to receive notifications instead of the Synapse object. The multi-watcher will then notify Synapse directly, *after resolving the different child watchers.*  

## Tests
- [x] Unit tests

## Reviewers
@bsherrod @austin-zhu

### DIFF
--- a/lib/synapse/service_watcher.rb
+++ b/lib/synapse/service_watcher.rb
@@ -11,15 +11,20 @@ module Synapse
         unless opts.has_key?('discovery') && opts['discovery'].has_key?('method')
 
       discovery_method = opts['discovery']['method']
+      return self.load_watcher(discovery_method, opts, synapse)
+    end
+
+    def self.load_watcher(discovery_method, opts, synapse)
       watcher = begin
-        method = discovery_method.downcase
-        require "synapse/service_watcher/#{method}"
-        # zookeeper_dns => ZookeeperDnsWatcher, ec2tag => Ec2tagWatcher, etc ...
-        method_class  = method.split('_').map{|x| x.capitalize}.join.concat('Watcher')
-        self.const_get("#{method_class}")
-      rescue Exception => e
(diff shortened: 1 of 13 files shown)


lib/synapse/service_watcher.rb
```diff
@@ -4,23 +4,28 @@
 module Synapse
   class ServiceWatcher
     # the method which actually dispatches watcher creation requests
-    def self.create(name, opts, synapse)
+    def self.create(name, opts, reconfigure_callback = nil, synapse)
```
austin-zhu: I don't see you have changed `ServiceWatcher.create` reference? Also, I think it is better to let the caller should pass in `synapse.reconfigure!` as callback explicitly.

panchr: That's correct, I did not change the reference because I thought it it was simpler to provide a default.

austin-zhu: I still prefer to pass in the value explicitly if you can. The default is not `nil` but something later changed inline. Also, you don't  even pass `nil` from `ServiceWatcher.create`. Though I think the method will still work because Ruby is smart and has a few tricks up its sleeve to help determine what method arguments are being used, I would suggest NOT relying on this. I am not sure whether all versions of Ruby will be that smart. It is conventional to place any default arguments at the end of an argument list when defining a method that takes in both required and default arguments.

panchr: Sounds good, I will do that and change the usage in the watchers

PR #314, opened 2020-03-03
Add resolver interface and base implementation [stacked on #311]

panchr:
_This PR is stacked on top of #311._

## Summary
A `Resolver` is responsible for providing the abstraction that multiple watchers are one: it decides how to merge the list of `backends` from each one. This PRs adds:
* the interface for the `Resolver`
* a base implementation
* documentation on how to create new `Resolver`s
* `MultiWatcher` loads the resolvers

## Testing
- [x] Unit tests

## Reviewers
@bsherrod @austin-zhu

### DIFF
--- a/lib/synapse/service_watcher/multi/multi.rb
+++ b/lib/synapse/service_watcher/multi/multi.rb
@@ -79,11 +79,11 @@ class Synapse::ServiceWatcher
     end
 
     def ping?
-      return @resolver.ping?
+      return @resolver.healthy?
     end
 
     def backends
-      return @resolver.backends
+      return @resolver.merged_backends
     end
 
     private
--- a/lib/synapse/service_watcher/multi/multi.rb
+++ b/lib/synapse/service_watcher/multi/multi.rb
@@ -53,7 +53,7 @@ class Synapse::ServiceWatcher
       end
 
       @resolver = Synapse::ServiceWatcher::Resolver.load_resolver(@discovery['resolver'],
-                                                                  @watchers.values)
+                                                                  @watchers)
     end
 
     def start
--- a/lib/synapse/service_watcher/multi/multi.rb
+++ b/lib/synapse/service_watcher/multi/multi.rb
@@ -1,4 +1,5 @@
 require "synapse/service_watcher/base/base"
+require "synapse/service_watcher/multi/resolver"
 require "synapse/log"
 require "synapse/statsd"
 
@@ -48,6 +49,9 @@ class Synapse::ServiceWatcher
 
         @watchers[watcher_name] = watcher
       end
+
+      @resolver = Synapse::ServiceWatcher::Resolver.load_resolver(@discovery['resolver'],
+                                                                  @watchers.values)
     end
 
     def start
@@ -57,6 +61,8 @@ class Synapse::ServiceWatcher
       @watchers.values.each do |w|
         w.start
       end
+
+      @resolver.start
     end
 
     def stop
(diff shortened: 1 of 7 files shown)


lib/synapse/service_watcher.rb
```diff
@@ -1,13 +1,17 @@
 require "synapse/log"
-require "synapse/service_watcher/base"
+require "synapse/service_watcher/base/base"
```
austin-zhu: NIT: for base one, just leave it under `synapse/service_watcher/base`? other derived watchers should have own folder

lib/synapse/service_watcher/multi/multi.rb
```diff
@@ -0,0 +1,113 @@
...
+
+    def start
+      log.info "synapse: starting multi watcher"
+      statsd_increment("synapse.watcher.multi.start")
+
+      @watchers.values.each do |w|
+        w.start
+      end
+
+      @resolver.start
```
austin-zhu: start resolver first or stop resolver first? the sequential should be mirrored

lib/synapse/service_watcher/multi/resolver/README.md
```diff
@@ -0,0 +1,56 @@
...
+	     # stop resolver
+	  end
+
+	  def backends
+	     # return a single list of backends
+	  end
+
+	  def ping?
+	     # return whether or not the watchers are healthy
+	  end
```
austin-zhu: find a better name of the method signature and make it more clear in the comments

lib/synapse/service_watcher/multi/resolver.rb
```diff
@@ -0,0 +1,18 @@
...
+    def self.load_resolver(opts, watchers)
+      raise ArgumentError, "resolver method not provided" unless opts.has_key?('method')
+      method = opts['method'].downcase
+
+      resolver = begin
+                   require "synapse/service_watcher/multi/resolver/#{method}"
+                   method_class  = method.split('_').map{|x| x.capitalize}.join.concat('Resolver')
+                   self.const_get("#{method_class}")
+                 rescue Exception => e
+                   raise ArgumentError, "Specified a resolver method of #{method}, which could not be found: #{e}"
```
austin-zhu: NIT: lower case first word

lib/synapse/service_watcher/multi/resolver/base.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      log.info "creating base resolver"
+
+      raise ArgumentError, "BaseResolver expects method to be base" unless opts['method'] == 'base'
+      raise ArgumentError, "no watchers provided" unless watchers.length > 0
+    end
+
+    # should be overridden in child classes
+    def start
+      log.info "starting base resolver"
```
austin-zhu: throw exception here if you want to force child class to implement it. Similar for other methods.

lib/synapse/service_watcher.rb
panchr: I avoided doing that because the `service_watcher.rb` loader requires a common path structure for every type, including the base watcher.

lib/synapse/service_watcher/multi/resolver/base.rb
panchr: The current `BaseWatcher` does not throw an exception; instead, the `BaseWatcher` is actually usable as-is; that allows you to run Synapse without any complicated setup (i.e. no Zookeeper or anything like that) by using the base watcher/resolver. I wanted to preserve that property

PR #315, opened 2020-03-05
Refactor Zookeeper watcher to support polling in future

panchr:
## Summary
Refactoring out some common code in order to support a separate `ZookeeperPollWatcher` after.
The common code includes:
* Setup/teardown of Zookeeper connections
* Reading child data (path-encoding support / fallback)

`ZookeeperPollWatcher` will be approximately:

```ruby
class ZookeeperPollWatcher < ZookeeperWatcher
   def start
      @stop_watcher = false
      @thread = nil

      zk_connect do
         @thread = Thread.new {
             until @stop_watcher
                 discover
                 sleep some_time
             end
         }
      end
   end

   def stop do
      @stop_watcher = true

      zk_teardown do
          @thread.stop unless @thread.nil?
      end
   end

   def discover
        zk_children = @zk.children(path)
        zk_children.each do |id|
          backend = read_child_data(id)
          new_backends << backend unless backend.nil?
        end

        new_config_for_generator = read_config_for_generator
        set_backends(new_backends, new_config_for_generator)
   end
end
```

## Tests
- [x] unit tests
- [x] manual testing
  * ran integration tests, including for network resilience
  * passes on all scenarios

## Reviewers
@austin-zhu @anson627

Tip: it's easier to review everything except for the commit `8d519b3` first. That commit just moves some functions around, so it's a lot of change but without much impact other than organization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ relevant routing component. For example if you want to only configure HAProxy an
 not NGINX for a particular service, you would pass ``disabled`` to the `nginx` section
 of that service's watcher config.
 
-* [`haproxy`](#haproxysvc): how will the haproxy section for this service be configured. If the corresponding `watcher` is defined to use `zookeeper` and the service publishes its `haproxy` configure on ZK, the `haproxy` hash can be filled/updated via data from the ZK node. 
+* [`haproxy`](#haproxysvc): how will the haproxy section for this service be configured. If the corresponding `watcher` is defined to use `zookeeper` and the service publishes its `haproxy` configure on ZK, the `haproxy` hash can be filled/updated via data from the ZK node.
 * [`nginx`](https://github.com/jolynch/synapse-nginx#service-watcher-config): how will the nginx section for this service be configured. **NOTE** to use this you must have the synapse-nginx [plugin](#plugins) installed.
 
 The services hash may contain the following additional keys:
@@ -259,7 +259,7 @@ Instead of setting Zookeeper watchers, it uses a long-polling method.
 It takes the following mandatory arguments:
 
 * `method`: zookeeper_poll
-* `polling_interval_sec`: the interval at which the watcher will poll Zookeeper.
+* `polling_interval_sec`: the interval at which the watcher will poll Zookeeper. Defaults to 60 seconds.
 
 Other than these two options, it takes the same options as the above ZookeeperWatcher.
 For all the required options, see above.
--- a/README.md
+++ b/README.md
@@ -251,6 +251,19 @@ If the `method` is `serverset` then we expect to find Finagle ServerSet
 (also used by [Aurora](https://github.com/apache/aurora/blob/master/docs/user-guide.md#service-discovery)) registrations with a `serviceEndpoint` and optionally one or more `additionalEndpoints`.
 The Synapse `name` will be automatically deduced from `shard` if present.
 
+##### Zookeeper Poll #####
+
+This watcher retrieves a list of servers and also service config data from zookeeper.
+Instead of setting Zookeeper watchers, it uses a long-polling method.
+
+It takes the following mandatory arguments:
+
+* `method`: zookeeper_poll
+* `polling_interval_sec`: the interval at which the watcher will poll Zookeeper.
+
+Other than these two options, it takes the same options as the above ZookeeperWatcher.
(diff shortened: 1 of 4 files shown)


lib/synapse/service_watcher/zookeeper_poll.rb
```diff
@@ -0,0 +1,77 @@
...
+          until @should_exit
+            now = Time.now
+            elapsed = now - last_run
+
+            if elapsed >= @poll_interval
+              last_run = now
+              discover
+            end
+
+            sleep 0.5
```
anson627: why sleep 0.5?

lib/synapse/service_watcher/zookeeper_poll.rb
```diff
@@ -0,0 +1,77 @@
...
+          # discover call of multiple Synapses. This helps to spread load.
+          # As long as the beginning is spread, the future discovers will also
+          # be spread.
+          last_run = Time.now - rand(@poll_interval)
+
+          until @should_exit
+            now = Time.now
+            elapsed = now - last_run
+
+            if elapsed >= @poll_interval
```
anson627: what happens if polling_interval_sec is not set? default value?

README.md
```diff
@@ -251,6 +251,19 @@ If the `method` is `serverset` then we expect to find Finagle ServerSet
...
 The Synapse `name` will be automatically deduced from `shard` if present.
 
+##### Zookeeper Poll #####
+
+This watcher retrieves a list of servers and also service config data from zookeeper.
+Instead of setting Zookeeper watchers, it uses a long-polling method.
+
+It takes the following mandatory arguments:
+
+* `method`: zookeeper_poll
```
anson627: why exposing "method"? Is it a flag to turn on/off polling? or you can switch btw poll and push?

lib/synapse/service_watcher/zookeeper_poll.rb
panchr: The idea here is to have an infinite loop in a background thread that can terminate quickly. We check whether or not to exit every iteration, so `sleep 0.5` caps the termination time to ~0.5 seconds.

If instead I do `sleep @poll_interval`, the thread could remain alive for a very long time potentially, depending on whatever the setting is.

README.md
panchr: The `method` is used by each watcher to identify it (and the loader uses this value to find it). I.e. the original `method` for Zookeeper is just `zookeeper`

lib/synapse/service_watcher/zookeeper_poll.rb
panchr: I was wary of setting a default value because I'm not sure what a good default would be--the acceptable load depends on how many Synapses you have running.

README.md
anson627: IC, so the `zookeeper` means `zookeeper_push`, and it is already configured everywhere.

lib/synapse/service_watcher/zookeeper_poll.rb
anson627: Just want to make sure, if the flag is not set, synapse shouldn't crash.

The default value can be any value sane and safe, in case some one put method as `zookeeper_poll` but doesn't set `polling_interval_sec`.

lib/synapse/service_watcher/zookeeper_poll.rb
anson627: make sense

README.md
panchr: Yup, that's correct

lib/synapse/service_watcher/zookeeper.rb
```diff
@@ -77,15 +93,92 @@ def ping?
...
+            log.warn "synapse: invalid discovery_jitter=#{@discovery['discovery_jitter']} for service:#{@name}"
+          end
+        end
+        watch
+        # Rediscover
+        discover
+      end
     end
 
+    protected
```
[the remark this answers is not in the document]
panchr: Nice catch!

I thought `protected` in Ruby was similar to C++, in that it allowed subclasses to call it. However, it seems that [`private` allows subclass access](https://stackoverflow.com/a/3534581/1730261) and `protected` is closer to `friend` in C++. I'll change it back to `private`

PR #321, opened 2020-04-06
Thread Safety

panchr:
## Summary
The new ZookeeperPoll watcher runs in a background thread and needs to communicate with the main thread (in order to tell it when to update). Currently, it just calls a callback which will execute in the background thread, and that has some thread safety concerns.

Also, the main thread will read the state of each watcher. The fixes for this are:

1. Hold a Synapse-level lock when reconfiguring.
2. Hold a watcher-level lock when getting/setting any watcher state.

In addition, it runs the primary Zookeeper watcher in a background thread. This allows retry logic to *not* block the main thread.

## Tests
- [x] unit tests
- [x] basic testing: running Synapse with multiple watchers
- [x] basic Zookeeper watcher still works
```bash
$ grep server haproxy.cfg
        timeout  server  1m
        server i-primary_127.0.0.1:1000 127.0.0.1:1000 id 2 cookie i-primary_127.0.0.1:1000 check inter 2s rise 3 fall 2
        server default1_localhost:8422 localhost:8422 id 1 cookie default1_localhost:8422 check inter 2s rise 3 fall 2

$ zookeepercli -servers localhost:2181 -c create /services/service1/0002.mango-test '{"host":"127.0.0.3","port":1003,"name":"i-primary3","labels":{"region":"us-east-1","az":"us-east-1c"}}'

$ grep server haproxy.cfg                                                                timeout  server  1m
        server i-primary3_127.0.0.3:1003 127.0.0.3:1003 id 1 cookie i-primary3_127.0.0.3:1003 check inter 2s rise 3 fall 2
        server i-primary_127.0.0.1:1000 127.0.0.1:1000 id 2 cookie i-primary_127.0.0.1:1000 check inter 2s rise 3 fall 2
        server default1_localhost:8422 localhost:8422 id 1 cookie default1_localhost:8422 check inter 2s rise 3 fall 2
```
- additional testing will be performed as part of mega-PR #311 

## Reviewers
@austin-zhu

### DIFF
--- a/lib/synapse/service_watcher/zookeeper/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper/zookeeper.rb
@@ -450,7 +450,7 @@ class Synapse::ServiceWatcher
       #   each key should be named by one of the available generators
       #   each value should be a hash (could be empty)
       decoded.collect.each do |generator_name, generator_config|
-        if !@synapse.available_generators.keys.include?(generator_name)
+        if !@available_generators.keys.include?(generator_name)
           log.warn "synapse: invalid generator name in ZK node at #{@discovery['path']}:" \
             " #{generator_name}"
           next
--- a/lib/synapse/service_watcher/zookeeper/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper/zookeeper.rb
@@ -1,4 +1,5 @@
 require "synapse/service_watcher/base/base"
+require 'synapse/atomic'
 
 require 'thread'
 require 'zk'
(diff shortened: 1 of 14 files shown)


lib/synapse/service_watcher/zookeeper/zookeeper.rb
```diff
@@ -54,29 +55,33 @@ def initialize(opts={}, synapse, reconfigure_callback)
...
-        with_retry(@retry_policy.merge({'retriable_errors' => ZK_RETRIABLE_ERRORS})) do |attempts|
-          unless @zk.exists?(@discovery['path'])
-            statsd_time('synapse.watcher.zk.create_path.elapsed_time', ["zk_cluster:#{@zk_cluster}", "service_name:#{@name}"]) do
-              log.info "synapse: zk create at #{@discovery['path']} for #{attempts} times"
-              create(@discovery['path'])
-            end
-          end
-        end
+        @thread = Thread.new {
+          start_discovery
```
austin-zhu, line 70: if the background thread is performing retry but there is a signal of `@should_exit`, it won't know until finishing retry, right? do you think it is better to move `@should_exit.get` into the retry block?

lib/synapse/service_watcher/zookeeper_poll/zookeeper_poll.rb
```diff
@@ -52,7 +52,7 @@ def stop
       zk_teardown do
         # Signal to the thread that it should exit, and then wait for it to
         # exit.
-        @should_exit = true
+        @should_exit.set(true) unless @should_exit.nil?
```
austin-zhu, line 55: why only checking the nil here?  Seems not necessary as no one could call `stop` until `initialize` returns

lib/synapse/service_watcher/zookeeper/zookeeper.rb
panchr: This is an existing problem which I wasn’t sure how to solve cleanly. The retry code is spread over a bunch of places and abstracted away into the `with_retry` function. So it’s hard to integrate the clean exit into there.

lib/synapse/service_watcher/zookeeper_poll/zookeeper_poll.rb
panchr: Good point! I added this in an earlier version where `@should_exit` was set in `start` instead of in `stop`.

PR #327, opened 2020-07-08
Set explicit zk session timeout

gmcatsf:
### Summary

Synapse does not specify explicit value for session timeout. By default ruby client uses RECEIVE_TIMEOUT=10000ms which leads to 20-sec session in zk server.

https://github.com/zk-ruby/zookeeper/blob/master/ext/c_zookeeper.rb#L20

The PR sets synapse session timeout to be 36-sec. 

### Test
- unit tests: bundle exec rspec
- To verify that zk session is set to 36-sec, manual integration test was executed with zk server: 
  - created an instance of client connecting to test zk server. 
  - used 4-letter commands to dump connections on zk server: `echo cons | nc localhost 2181 > cons.txt`
  - verified the output that connection had 36-sec session with to=36001: 
`/172.18.26.75:53573[1](queued=0,recved=2,sent=2,sid=0x66055392eb24038e,lop=PING,est=1594240596217,to=36001,lcxid=0x0,lzxid=0xffffffffffffffff,lresp=22948687231,llat=1,minlat=0,avglat=2,maxlat=3)`

## Reviewers
@panchr 
@austin-zhu
@Jason-Jian


gmcatsf:
updated version to 0.18.3

PR #331, opened 2020-08-18
Reduce zk calls with enhanced zk event processing

gmcatsf:
### Summary

Current ZookeeperWatcher's watch_callback accepts zk event as parameter but ignores it during processing. Every time a zk event is received, watch_callback does all three things regardless of which zk event:

- calling zk.exists on service znode
- calling zk.children on service znode
- calling zk.get on service znode.

But each zk event always contains single change(e.g. child and data event are mutually exclusive) and performing all three calls above are not necessary. For example below are a list of methods supported on zk event:

- node_event? : true for node event and false for session event
- node_child? : true if watched znode children have changed.
- node_chanaged? : true if watched znode data has changed.
- node_created? : true if watched znode was created 
- node_deleted? : true if watched znode was deleted.

This PR adds a few boolean checks to handle children change and data change separately, and should reduce three zk calls to one in most cases since children change is the most frequent ones.

During watch startup, the three calls will still be executed to read data and set watches in zk.

### Test
- Added new rspec tests to test nil/child/changed events: bundle exec rspec
- Ran ZookeeperWatcher against zookeeper-test and checked event handling with manual zk data change.

### Reviewer
@airbnb/traffic
@panchr 
@austin-zhu 
@Jason-Jian

### DIFF
--- a/lib/synapse/service_watcher/zookeeper/zookeeper.rb
+++ b/lib/synapse/service_watcher/zookeeper/zookeeper.rb
@@ -113,38 +113,55 @@ class Synapse::ServiceWatcher
     end
 
     # find the current backends at the discovery path
-    def discover(zookeeper_opts={:watch => true})
+    def discover(zookeeper_opts={:watch => true}, zk_event=nil)
       statsd_increment('synapse.watcher.zk.discovery', ["zk_cluster:#{@zk_cluster}", "zk_path:#{@discovery['path']}", "service_name:#{@name}"])
       statsd_time('synapse.watcher.zk.discovery.elapsed_time', ["zk_cluster:#{@zk_cluster}", "zk_path:#{@discovery['path']}", "service_name:#{@name}"]) do
         log.info "synapse: discovering backends for service #{@name}"
 
-        new_backends = []
-        zk_children = with_retry(@retry_policy.merge({'retriable_errors' => ZK_RETRIABLE_ERRORS})) do |attempts|
-            statsd_time('synapse.watcher.zk.children.elapsed_time', ["zk_cluster:#{@zk_cluster}", "service_name:#{@name}"]) do
-              log.info "synapse: zk list children at #{@discovery['path']} for #{attempts} times"
-              begin
-                @zk.children(@discovery['path'], zookeeper_opts)
-              rescue ZK::Exceptions::NoNode
-                log.error "synapse: zk list children failed with no node"
-                statsd_increment('synapse.watcher.zk.children.failed', ["zk_cluster:#{@zk_cluster}", "service_name:#{@name}"])
-                []
+        # There are three possible cases, with each case exclusive to others.
+        # 1. No zk event, read data with both data/child watches registered
+        # 2. Registered child watch fired, need_get_children is set to true and children are read from zk with new child watch registered.
+        # 3. Registered data watch fired, need_read_config is set to true and znode is read from zk with new data watch registered.
+        need_get_children = zk_event.nil? || zk_event.node_child?
+        need_read_config = zk_event.nil? || zk_event.node_changed? || zk_event.node_created? || zk_event.node_deleted?
+
+        log.info "synapse: read children: #{need_get_children}, read config: #{need_read_config}"
+
+        new_backends = nil
+        if need_get_children
+          new_backends = []
+          zk_children = with_retry(@retry_policy.merge({'retriable_errors' => ZK_RETRIABLE_ERRORS})) do |attempts|
+              statsd_time('synapse.watcher.zk.children.elapsed_time', ["zk_cluster:#{@zk_cluster}", "service_name:#{@name}"]) do
+                log.info "synapse: zk list children at #{@discovery['path']} for #{attempts} times"
+                begin
+                  @zk.children(@discovery['path'], zookeeper_opts)
+                rescue ZK::Exceptions::NoNode
+                  log.error "synapse: zk list children failed with no node"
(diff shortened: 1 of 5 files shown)


lib/synapse/service_watcher/zookeeper/zookeeper.rb
```diff
@@ -160,11 +177,19 @@ def watch
...
 
-        # Verify that we actually set up the watcher.
-        existed = with_retry(@retry_policy.merge({'retriable_errors' => ZK_RETRIABLE_ERRORS})) do |attempts|
-          log.info "synapse: zk exists at #{@discovery['path']} for #{attempts} times"
-          @zk.exists?(@discovery['path'], :watch => true)
+        # There are two cases when new watch needs to be created in zk, otherwise watch already exists.
+        # 1. Create watch in zk when zk event is nil
+        # 2. Create watch again when current event is triggered by existing watch. 
+        existed = true
+        need_exists_watch = zk_event.nil? or zk_event.node_changed? or zk_event.node_created? or zk_event.node_deleted?
```
panchr, line 184: nit: use `||` instead of `or` for consistency

lib/synapse/service_watcher/zookeeper/zookeeper.rb
```diff
@@ -191,6 +216,7 @@ def watcher_callback
         # We call exists(watch), children(discover) and get(discover) with (:watch=>true)
         # This re-initializes callbacks just before get scan. So we do not miss any updates by sleeping.
         #
+        puts 'callback in invoked'
```
panchr, line 219: I think this was a debugging comment? please remove if so, or switch to `log.debug`

gmcatsf: Good catch!

lib/synapse/service_watcher/zookeeper/zookeeper.rb
gmcatsf: will update

gmcatsf:
@austin-zhu could you please help check the code change?

lib/synapse/service_watcher/zookeeper/zookeeper.rb
```diff
@@ -113,38 +113,55 @@ def start_discovery
...
-              rescue ZK::Exceptions::NoNode
-                log.error "synapse: zk list children failed with no node"
-                statsd_increment('synapse.watcher.zk.children.failed', ["zk_cluster:#{@zk_cluster}", "service_name:#{@name}"])
-                []
+        # There are three possible cases, with each case exclusive to others.
+        # 1. No zk event, read data with both data/child watches registered
+        # 2. Registered child watch fired, need_get_children is set to true and children are read from zk with new child watch registered.
+        # 3. Registered data watch fired, need_read_config is set to true and znode is read from zk with new data watch registered.
+        need_get_children = zk_event.nil? || zk_event.node_child?
+        need_read_config = zk_event.nil? || zk_event.node_changed? || zk_event.node_created? || zk_event.node_deleted?
```
austin-zhu, line 126: NIT: under some cases, like `node_deleted`, it is actually not to read the config. Suggest changing it to `need_update_config`.

lib/synapse/service_watcher/zookeeper/zookeeper.rb
```diff
@@ -154,11 +171,19 @@ def watch
...
 
-        # Verify that we actually set up the watcher.
-        existed = with_retry(@retry_policy.merge({'retriable_errors' => ZK_RETRIABLE_ERRORS})) do |attempts|
-          log.info "synapse: zk exists at #{@discovery['path']} for #{attempts} times"
-          @zk.exists?(@discovery['path'], :watch => true)
+        # There are two cases when new watch needs to be created in zk, otherwise watch already exists.
+        # 1. Create watch in zk when zk event is nil
+        # 2. Create watch again when current event is triggered by existing watch. 
+        existed = true
+        need_exists_watch = zk_event.nil? || zk_event.node_changed? || zk_event.node_created? || zk_event.node_deleted?
```
austin-zhu, line 178: why does `zk_event.node_changed?` also need to set another watch?

lib/synapse/service_watcher/zookeeper/zookeeper.rb
```diff
@@ -113,38 +113,55 @@ def start_discovery
     end
 
     # find the current backends at the discovery path
-    def discover(zookeeper_opts={:watch => true})
+    def discover(zookeeper_opts={:watch => true}, zk_event=nil)
```
austin-zhu, line 116: @gmcatsf just to confirm, when reconnected but session timeout, it will pass `nil` event so that all watches will be set?

austin-zhu (approved):

lib/synapse/service_watcher/zookeeper/zookeeper.rb
gmcatsf: Just to clarify, this method only handles registered node events and is not getting called when connection/session events occur.

In case of reconnected but session timeout, current event handling logic is to set @zk to nil, and then synapse will raise error and restart ZookeeperWatcher. At that time your statement is true, i.e. it passes nil event so all watches will be set.

lib/synapse/service_watcher/zookeeper/zookeeper.rb
gmcatsf: ZK has two types of watches, one is data watch on znode , the other is child watch on znode children. The watch has nothing to do with type of calls is made, exists() or getData(), that means this watch will be triggered when the znode is changed not deleted, so it needs to set watch again.

